PR #4, opened 2016-06-27
Added get all items method

atakangktepe:
Issue: #5


sindresorhus:
You can already get all the config using `config.store`.

Issue #2, opened 2016-06-22
Build not passing - silent fail? 


kasperlewau:
I hate to be this guy *(chill, it's 0.10.0...)* but, you've got an [.only](https://github.com/sindresorhus/conf/blob/master/test.js#L144) that's reporting a non-passing build as passing on Node@5.11.1 & Node@6.2.0 with a clean `npm i`. 

![failing](https://i.imgur.com/4mzDLK4.png)

Builds on 
* Travis (https://travis-ci.org/sindresorhus/conf/jobs/139352324#L653)
* AppVeyor (https://ci.appveyor.com/project/sindresorhus/conf/build/6/job/ccvy2qihrr40tqh8#L770)

--- 

I'd submit a PR, but my head is not where it's supposed to be right now and duty calls (*laundry...*). Just though I'd let you know. Cheers! 🔨

sindresorhus:
Ugh. Thanks for reporting :) Luckily, I only had a couple of incorrect tests. Code was working perfectly.

`.only` is so dangerous! I actually have XO, which bundles the https://github.com/avajs/eslint-plugin-ava/blob/master/docs/rules/no-only-test.md rule, which should have caught this. @jfmengels Any idea why it didn't?

jfmengels:
Hmm, no, sounds really odd. I'll try to look into it later, but it is odd indeed.

jfmengels:
`import {serial as test} from 'ava';` it's this part that we don't recognize. Do we want to handle this case or does that seem like a weird use of AVA? IMO, it kind of does :D, but we should probably try to handle it.

Didn't know you could destructure things like `serial` and that it was the same thing as `test.serial`. We actually don't even handle anything when AVA is not imported as `test`. Plenty of room for improvement here.

sindresorhus:
@jfmengels I opened an issue: https://github.com/avajs/eslint-plugin-ava/issues/121

PR #6, opened 2016-07-11
Breaks if no package.json is present

JimmyBoh:
If there is no `package.json` available then the path is `null` and you can't stop it from breaking.

Even if you manually provide the `projectName` in options it will still break. This adds a few simple checks to prevent it from blowing up when you do provide the name.

I've also added an appropriate test, which properly fails when run without the changes to `index.js` and correctly passes once added.


index.js
```diff
@@ -15,9 +15,13 @@ const parentDir = path.dirname(module.parent.filename);
...
 
-		if (!opts.projectName) {
+		// If the package.json was not found, avoid breaking with `require(null)`.
+		if (pkgPath) {
+			opts = Object.assign({projectName: require(pkgPath).name}, opts);
+		}
+
+		if (!opts || !opts.projectName) {
 			throw new Error('Project name could not be inferred. Please specify the `projectName` option.');
 		}
```
sindresorhus: I think I would rather do:

```js
const pkgPath = pkgUp.sync(parentDir);

opts = Object.assign({
	// if the package.json was not found, avoid breaking with `require(null)`.
	projectName: pkgPath && require(pkgPath).name
}, opts);

if (!opts.projectName) {
	throw new Error('Project name could not be inferred. Please specify the `projectName` option.');
}
```

sindresorhus:
Looks good.

Just curious, why is there no package.json available?

index.js
JimmyBoh: Tons of ways to handle this null value. I will update it to match yours, nice and clean.

sindresorhus: Yup, totally subjective. Both work equally fine :)

sindresorhus:
Can you fix the merge conflict? Sorry, I had a local fix that I pushed by accident now.

JimmyBoh:
This is limited to select installs of Electron apps (depending on how they are packaged) but I use [this one](https://github.com/szwacz/electron-boilerplate) by @szwacz. I'm not sure if there are any other downsides, but I figured an unavoidable exception like this could be a problem for others as well.

JimmyBoh:
And no worries about the merge conflict. I'll fix that once I get home (done at the office for today).

JimmyBoh:
I just noticed the test you added [still has `.only` on it](https://github.com/sindresorhus/conf/commit/622f648fed33166e331e8859c8320e395c6e5fad#diff-1dd241c4cd3fd1dd89c570cee98b79ddR150). Do you want to fix that before I merge? Or you want me to take care of that right now?

sindresorhus:
[Ugh, not again.](https://github.com/sindresorhus/conf/issues/2#issuecomment-227813534) Fixed in master. Thanks for noticing.

JimmyBoh:
No problem, happens to the best of us!

sindresorhus:
Thank you :)

JimmyBoh:
My pleasure!

Quick (potentially silly) question: This does not say "merged", but instead closed with a commit that "I committed with you". Is that the result of Github squashing the commits? Or something you did to keep things clean? Next time should I have squashed my commits first?

SamVerschueren:
GitHub now has a button to squash and merge a PR so no need in doing that yourself, although you can off course :).

Not sure if Sindre uses the button though. Before there was a button to squash, he did everything "manually".

JimmyBoh:
Hey @SamVerschueren, I've seen that but have yet to use it. I'm guessing he did things manually, hence the "closed" status as opposed to Github's "merged" status. Thanks for the insight!

JimmyBoh:
Oh, and @sindresorhus, may I recommend a bump to [`electron-config`](https://github.com/sindresorhus/electron-config), so that it can have the latest updates as well (`"conf": "^0.10.0"` restricts it to `0.10.*`).

Thank you!

sindresorhus:
> I've seen that but have yet to use it. I'm guessing he did things manually, hence the "closed" status as opposed to Github's "merged" status. Thanks for the insight!

Yeah. I did it manually as I wanted to tweak something on merge. I could probably have used squash and then force pushed, but old habits die hard.

> may I recommend a bump to electron-config

Done :)

JimmyBoh:
Thank You!

:]

PR #9, opened 2016-09-07
improve set input detection - fixes #8

### LINKED ISSUE #8: `.set(object)` detection might be wrong
Currently, this is the implementation of `.set(key, val)` which supports `.set(object)` as well.

```js
set(key, val) {
	const store = this.store;

	if (val === undefined) {
		Object.keys(key).forEach(k => {
			dotProp.set(store, k, key[k]);
		});
	} else {
		dotProp.set(store, key, val);
	}

	this.store = store;
}
```

If someone calls `set('foo')` or `set('foo', undefined)`, `Object.keys('foo')` returns `[0, 1, 2]` and the store will look like this 

```
{
    '0': 'f', 
    '1': 'o', 
    '2': 'o'
}
```

I believe we should add extra checks to the `set` method. But the first question to answer is, is `string` the only valid input type for `key`?

1. Test that `key` is of type `string` or `object`.
2. Test if `key` is of type `object` AND `value` is `undefined`.
  - We could ignore `value` in this case
  - We could throw an error indicating that it's not possible to provide a `value` if `key` is of type `object`

SamVerschueren:
Things I missed?

Should we also check `if (typeof key === 'string' && val === undefined)` and throw an error? Or should we allow to overwrite a value with `undefined`? Seems a valid use case...

### DIFF
--- a/index.js
+++ b/index.js
@@ -40,9 +40,13 @@ class Conf {
 		return dotProp.get(this.store, key);
 	}
 	set(key, val) {
+		if (typeof key !== 'string' && typeof key !== 'object') {
+			throw new TypeError(`Expected \`key\` to be of type \`string\` or \`object\`, got ${typeof key}`);
+		}
+
 		const store = this.store;
 
-		if (val === undefined) {
+		if (typeof key === 'object') {
 			Object.keys(key).forEach(k => {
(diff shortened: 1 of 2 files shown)


sindresorhus:
`undefined` should be allowed, just like it's allowed in a `Map`.

PR #14, opened 2017-02-02
Add support for providing a default value with getter

Ignigena:
The `4.x` version of `dot-prop` includes the ability to pass a default with the `.get()` function which will be returned if the property is not defined. I am using your module `electron-config` and would like this functionality included for the added convenience of setting a default configuration state when one has not been saved for the user.

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,8 +36,8 @@ class Conf {
 		this.path = path.resolve(opts.cwd, `${opts.configName}.json`);
 		this.store = Object.assign(obj(), opts.defaults, this.store);
 	}
-	get(key) {
-		return dotProp.get(this.store, key);
+	get(key, defaultValue) {
+		return dotProp.get(this.store, key, defaultValue);
 	}
 	set(key, val) {
 		if (typeof key !== 'string' && typeof key !== 'object') {
--- a/readme.md
+++ b/readme.md
@@ -90,9 +90,9 @@ Set an item.
 
 Set multiple items at once.
 
-#### .get(key)
+#### .get(key, [defaultValue])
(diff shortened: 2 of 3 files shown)


sindresorhus:
Sure. Looks like a nice shortcut.

readme.md
```diff
@@ -90,7 +90,7 @@ Set an item.
 
 Set multiple items at once.
 
-#### .get(key)
+#### .get(key, [value])
```
sindresorhus: value => defaultValue

index.js
```diff
@@ -36,8 +36,8 @@ class Conf {
 		this.path = path.resolve(opts.cwd, `${opts.configName}.json`);
 		this.store = Object.assign(obj(), opts.defaults, this.store);
 	}
-	get(key) {
-		return dotProp.get(this.store, key);
+	get(key, val) {
```
sindresorhus: val => defaultValue

readme.md
Ignigena: Done!

index.js
Ignigena: Done!

readme.md
```diff
@@ -90,7 +90,7 @@ Set an item.
 
 Set multiple items at once.
 
-#### .get(key)
+#### .get(key, [defaultValue])
```
SamVerschueren: Should we document this?

> Get an item or `defaultValue` if the item does not exist.

Ignigena: Good point. Updated!

sindresorhus:
Thanks :)

PR #16, opened 2017-03-22
Fix #11 by falling back to cwd

### LINKED ISSUE #11: Support running it in a REPL
See: https://github.com/sindresorhus/conf/issues/10#issuecomment-249320242

florianb:
The usage of conf fails in the REPL environment because the `module.filename` is `null`. In order to play around with the current module's config, i added the search & use of a local `package.json` if it exists.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,16 +13,10 @@ delete require.cache[__filename];
 class Conf {
 	constructor(opts) {
 		let pkgPath;
-		// Fallback to current package.json if module.parent.filename is null (f.e. in REPL)
 		if (module.parent.filename) {
 			pkgPath = pkgUp.sync(path.dirname(module.parent.filename));
 		} else {
-			pkgPath = path.join(__dirname, 'package.json');
-			try {
-				fs.accessSync(pkgPath, fs.constants.F_OK | fs.constants.R_OK);
-			} catch (err) {
-				pkgPath = null;
-			}
+			pkgPath = pkgUp.sync(process.cwd());
 		}
 
 		opts = Object.assign({
--- a/index.js
+++ b/index.js
@@ -10,11 +10,20 @@ const obj = () => Object.create(null);
 
 // Prevent caching of this module so module.parent is always accurate
 delete require.cache[__filename];
-const parentDir = path.dirname(module.parent.filename);
-
 class Conf {
 	constructor(opts) {
-		const pkgPath = pkgUp.sync(parentDir);
+		let pkgPath;
+		// Fallback to current package.json if module.parent.filename is null (f.e. in REPL)
+		if (module.parent.filename) {
+			pkgPath = pkgUp.sync(path.dirname(module.parent.filename));
+		} else {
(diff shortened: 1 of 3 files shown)


florianb:
The current implementation unfortunately doesn't work in the test-environment (but in the REPL) - i couldn't figure out why, yet..

florianb:
Ready to :shipit:

test.js
```diff
@@ -193,3 +193,15 @@ test('handle `cwd` being set and `projectName` not being set', t => {
 	del.sync(conf.path, {force: true});
 	pkgUp.sync = pkgUpSyncOrig;
 });
+
+// See https://github.com/sindresorhus/conf/issues/11
+test('fallback to cwd if `module.filename` is `null`', t => {
+	module.filename = null;
```
sindresorhus: You need to restore this after the test so it doesn't affect the other tests.

index.js
```diff
@@ -10,11 +10,9 @@ const obj = () => Object.create(null);
 
 // Prevent caching of this module so module.parent is always accurate
 delete require.cache[__filename];
-const parentDir = path.dirname(module.parent.filename);
```
sindresorhus: I would prefer to still keep this here and add the `.` here instead.

You can use https://github.com/sindresorhus/require-uncached to be able to test it.

test.js
florianb: Of course :man_facepalming:

index.js
florianb: 🤔 i was removing it since it is unused and it's parameter is finally falling back to `.`. It is - of course - no problem to create that const again. But it would require some kind of block statement to avoid having a null-value being evaluated. In addition to that, i'd like to avoid having `parentDir` containing two different values (either the parent directory or a reference to the current working directory).

I could imagine to solve it like this:
```
const parentDir = module.parent.filename === null ? null : path.dirname(module.parent.filename);
...
pkgUp.sync(parentDir || '.');
```

What would you say (I'd like to avoid discussing this after commits)?

sindresorhus: By "keep this here", I meant instead of the other place you moved it to. Line 13 becomes:

```js
const parentDir = path.dirname(module.parent.filename || '.');
```

No reason to resolve it in the class when it can be done once on line 13.

florianb: Ahh - okay. Nevermind then.

florianb: Done - btw `require-uncached` didn't do the trick, because the modules-parent was then pointing to the uncached module. I made use of your neat `clear-require`. 🍻

test.js
```diff
@@ -195,15 +195,15 @@ test('handle `cwd` being set and `projectName` not being set', t => {
 	pkgUp.sync = pkgUpSyncOrig;
 });
 
-// See https://github.com/sindresorhus/conf/issues/11
+// See #11
```
florianb: I guess i will propose a 🦄  rule for that

sindresorhus:
Thank you @florianb :)

PR #21, opened 2017-08-26
Add encryption support

callodacity:
Initially got the idea from [here](https://github.com/electron-userland/electron-json-storage/issues/23) so I thought I might add it to my preferred config/store module 👌

### DIFF
--- a/index.js
+++ b/index.js
@@ -103,7 +103,7 @@ class Conf {
 		let data = JSON.stringify(val, null, '\t');
 		if (this.encryptionKey) {
 			const cipher = crypto.createCipher('aes-256-cbc', this.encryptionKey);
-			data = Buffer.concat([cipher.update(data), cipher.final()]);
+			data = Buffer.concat([cipher.update(Buffer.from(data)), cipher.final()]);
 		}
 
 		writeFileAtomic.sync(this.path, data);
--- a/index.js
+++ b/index.js
@@ -75,12 +75,11 @@ class Conf {
 	}
 	get store() {
 		try {
-			let data = fs.readFileSync(this.path, 'utf8');
+			let data = fs.readFileSync(this.path, this.encryptionKey ? null : 'utf8');
 
 			if (this.encryptionKey) {
 				const decipher = crypto.createDecipher('aes-256-cbc', this.encryptionKey);
-				data = decipher.update(data, 'hex', 'utf8');
-				data += decipher.final('utf8');
+				data = Buffer.concat([decipher.update(data), decipher.final()]);
 			}
 
 			return Object.assign(obj(), JSON.parse(data));
@@ -104,8 +103,7 @@ class Conf {
 		let data = JSON.stringify(val, null, '\t');
 		if (this.encryptionKey) {
 			const cipher = crypto.createCipher('aes-256-cbc', this.encryptionKey);
-			data = cipher.update(data, 'utf8', 'hex');
-			data += cipher.final('hex');
+			data = Buffer.concat([cipher.update(data), cipher.final()]);
 		}
 
 		writeFileAtomic.sync(this.path, data);
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict';
 const fs = require('fs');
 const path = require('path');
+const crypto = require('crypto');
 const dotProp = require('dot-prop');
 const makeDir = require('make-dir');
 const pkgUp = require('pkg-up');
@@ -33,6 +34,7 @@ class Conf {
 			opts.cwd = envPaths(opts.projectName).config;
 		}
 
+		this.cryptoKey = opts.cryptoKey || null;
 		this.path = path.resolve(opts.cwd, `${opts.configName}.json`);
 		this.store = Object.assign(obj(), opts.defaults, this.store);
 	}
@@ -72,7 +74,16 @@ class Conf {
 	}
 	get store() {
 		try {
-			return Object.assign(obj(), JSON.parse(fs.readFileSync(this.path, 'utf8')));
+			let data = fs.readFileSync(this.path, 'utf8');
+
+			// If `options.cryptoKey` is given, decrypt data.
+			if (this.cryptoKey) {
+				const decipher = crypto.createDecipher('aes-256-cbc', this.cryptoKey);
+				data = decipher.update(data, 'hex', 'utf8');
+				data += decipher.final('utf8');
+			}
+
+			return Object.assign(obj(), JSON.parse(data));
 		} catch (err) {
 			if (err.code === 'ENOENT') {
 				makeDir.sync(path.dirname(this.path));
@@ -90,7 +101,16 @@ class Conf {
 		// Ensure the directory exists as it could have been deleted in the meantime
 		makeDir.sync(path.dirname(this.path));
 
-		fs.writeFileSync(this.path, JSON.stringify(val, null, '\t'));
+		let data = JSON.stringify(val, null, '\t');
+
+		// If `options.cryptoKey` is given, encrypt data.
+		if (this.cryptoKey) {
+			const cipher = crypto.createCipher('aes-256-cbc', this.cryptoKey);
+			data = cipher.update(data, 'utf8', 'hex');
+			data += cipher.final('hex');
+		}
+
+		fs.writeFileSync(this.path, data);
 	}
 	// TODO: Use `Object.entries()` here at some point
 	* [Symbol.iterator]() {
--- a/index.js
+++ b/index.js
@@ -102,7 +102,6 @@ class Conf {
 
 		let data = JSON.stringify(val, null, '\t');
 
-		// If `options.encryptionKey` is given, encrypt data.
 		if (this.encryptionKey) {
 			const cipher = crypto.createCipher('aes-256-cbc', this.encryptionKey);
 			data = cipher.update(data, 'utf8', 'hex');
--- a/index.js
+++ b/index.js
@@ -34,7 +34,7 @@ class Conf {
 			opts.cwd = envPaths(opts.projectName).config;
 		}
 
-		this.cryptoKey = opts.cryptoKey || null;
+		this.encryptionKey = opts.encryptionKey;
 		this.path = path.resolve(opts.cwd, `${opts.configName}.json`);
 		this.store = Object.assign(obj(), opts.defaults, this.store);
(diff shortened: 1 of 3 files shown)


callodacity:
If you're happy with the addition I'll make changes to the readme as well 👍

sindresorhus:
If someone has access to the config file, they also have access to the encryption key, right? Can you elaborate on how you would use this?

index.js
```diff
@@ -72,7 +74,16 @@ class Conf {
 	}
 	get store() {
 		try {
-			return Object.assign(obj(), JSON.parse(fs.readFileSync(this.path, 'utf8')));
+			let data = fs.readFileSync(this.path, 'utf8');
+
+			// If `options.cryptoKey` is given, decrypt data.
+			if (this.cryptoKey) {
+				const decipher = crypto.createDecipher('aes-256-cbc', this.cryptoKey);
```
sindresorhus: Why did you pick `'aes-256-cbc'`?

index.js
```diff
@@ -33,6 +34,7 @@ class Conf {
 			opts.cwd = envPaths(opts.projectName).config;
 		}
 
+		this.cryptoKey = opts.cryptoKey || null;
```
sindresorhus: `cryptoKey` => `encryptionKey`

And the `null` fallback is moot.

index.js
```diff
@@ -72,7 +74,16 @@ class Conf {
 	}
 	get store() {
 		try {
-			return Object.assign(obj(), JSON.parse(fs.readFileSync(this.path, 'utf8')));
+			let data = fs.readFileSync(this.path, 'utf8');
+
+			// If `options.cryptoKey` is given, decrypt data.
```
sindresorhus: Drop the comment, it doesn't explain anything more than what the code does.

test.js
```diff
@@ -209,3 +210,18 @@ test('fallback to cwd if `module.filename` is `null`', t => {
 	module.filename = preservedFilename;
 	del.sync(conf.path, {force: true});
 });
+
+// Test encryption.
```
sindresorhus: Moot comment

test.js
```diff
@@ -209,3 +210,18 @@ test('fallback to cwd if `module.filename` is `null`', t => {
...
+test('.get() (encrypted)', t => {
```
sindresorhus: `'.get() (encrypted)'` => `'.get() - encryption'`

sindresorhus: Actually, I don't think you need two tests for this almost doing the same thing, just test `get` and `set` in the same test.

callodacity:
How would they have access to the encryption key if they have access to the config file?  

The config file will just be an aes encrypted file - you can't get any information out of it without the key. Storing the key would be something left up to the user.

index.js
callodacity: Just one that I was familiar with - we could include the encryption method as an option if you'd like. 

https://nodejs.org/api/crypto.html#crypto_crypto_createcipher_algorithm_password

sindresorhus:
I'm just trying to understand for what and how you would use this. The main use-case for this module is for CLI tools and apps (Like Electron). Both of which it makes no sense to encrypt the key as the encryption key would have to be stored in the source, some config file, or environment variables, all of which the user has access to anyway.

callodacity:
That's all definitely true, a user could read through all the electron app's source code and find the encryption key somewhere. It [seemed](https://github.com/electron-userland/electron-json-storage/issues/23) like others were interested in it too.

If security really were a concern, than an electron app is not the way to go. The use case I was just thinking was just to simply obfuscate the config file of an electron app so if a user were perusing through the app's data directory and saw a simple json file, they may be tempted to change it or add in random values. If it were encrypted, it may be a deterrent.  
It also verifies the integrity of the config - if someone changes a json key or value, nothing will happen, if they change a char in an encrypted file it won't be decrypted properly.

I really just whipped this up to see if anyone else had any similar use cases, and to see if anyone else was really interested in it (I don't mind if this is merged or not, so don't feel like I'll be disappointed if you say no it it 😝 haha).

sindresorhus:
It's a simple change, so I'll take it, but it needs docs and a clear note of it not being secure unless you don't include the key, but that it can be useful for obscurity as you've described above.

callodacity:
No worries, I'll push a change to the docs as well 👍

test.js
```diff
@@ -10,6 +10,7 @@ const fixture = '🦄';
 
 test.beforeEach(t => {
 	t.context.conf = new Conf({cwd: tempy.directory()});
+	t.context.cryptoConf = new Conf({cwd: tempy.directory(), cryptoKey: 'abc123'});
```
sindresorhus: The instance should just be initiated in the test body, not here, since it's not shared anywhere else anymore.

index.js
```diff
@@ -90,7 +100,15 @@ class Conf {
 		// Ensure the directory exists as it could have been deleted in the meantime
 		makeDir.sync(path.dirname(this.path));
 
-		fs.writeFileSync(this.path, JSON.stringify(val, null, '\t'));
+		let data = JSON.stringify(val, null, '\t');
+
+		if (this.encryptionKey) {
+			const cipher = crypto.createCipher('aes-256-cbc', this.encryptionKey);
+			data = cipher.update(data, 'utf8', 'hex');
+			data += cipher.final('hex');
```
sindresorhus: Wouldn't it be better to write it as a buffer? So without the `'hex'` argument. It would be faster to both read and write. You would then have to conditionally read as a buffer instead of `utf8` too though.

callodacity: I was trying to mimic the current functionality as much as possible - just modify the `data` variable but keep the same type.

index.js
callodacity: In regards to this, we could change the options from:

    new Conf({ encryptionKey: 'my_key' });

To something like:

    new Conf({ encryption: { key: 'my_key', algorithm: 'aes-256-cdc' });

That way the user can choose the algorithm. Since [the algorithm is dependent on OpenSSL](https://nodejs.org/api/crypto.html#crypto_crypto_createcipher_algorithm_password) it may be useful for some systems whose OpenSSL doesn't have `aes-256-cbc`.

---

And if `encryption` is just a string, then fallback to just using that as the encryption key with a sensible default.

readme.md
```diff
@@ -78,6 +78,17 @@ Overrides `projectName`.
...
+
+Type: `string | Buffer | TypedArray | DataView`<br>
+Default: `undefined`
+
+Note that this **is not intended for security purposes**, since the encryption key would be easily found inside a plain-text NodeJS application.
+
+Its main use is for obscurity: if a user was looking through the user config directory and found the config file, since it's a simple json file they may be tempted to modify/add/remove a key/value pair. By providing an encryption key `Conf` will obfuscate the file which should hopefully deter any users from doing so.
+
+It also has the added bonus that it will maintain a config file's integrity: if the file is changed in any way then decryption will not work, in which case the store will just reset back to its default state.
+
```
sindresorhus: Can you include that it's encrypted with `aes-256-cbc` and linkfy the name with a resource about this encryption.

callodacity: I was thinking about this, and I think we should change the options from:

    new Conf({ encryptionKey: 'my_key' });

To something like:

    new Conf({ encryption: { key: 'my_key', algorithm: 'aes-256-cdc' });

That way the user can choose the algorithm. Since [the algorithm is dependent on OpenSSL](https://nodejs.org/api/crypto.html#crypto_crypto_createcipher_algorithm_password) it may be useful for some systems whose OpenSSL doesn't have `aes-256-cbc`. (I'm not familiar with different installations of OpenSSL, so I don't know if they do or don't have this style of encryption).

---

And if `encryption` is just a string, then fallback to just using that as the encryption key with a sensible default.

sindresorhus: Nah, I prefer to keep it simple.

I always keep this in the back of my head when designing APIs: 😛

horrible-ui

callodacity:
@sindresorhus looks like using buffers isn't very nice to `Node 4`. It continues to be decrypted incorrectly - I can't figure out exactly why, though. I think perhaps using the `'hex'` approach (or any other encoded format, `base64`, etc.) is going to be more consistent.

sindresorhus:
Very nice work. Thanks for doing this :)

callodacity:
I just realised that there may be a confusing case when a user begins to use encryption for the first time. Consider this scenario:

```js
let conf = new Conf();
conf.set('foo', 'bar');
// ...
```

And then, when they upgrade their app/module, and they implement encryption:

```js
let conf = new Conf({ encryptionKey: 'my_key' }); // throws!
```

It throws because `Conf` sees an encryption key, and begins to decrypt the data - however, the data isn't yet encrypted because the user has only just enabled encryption.

A workaround could be something like:

```js
// This only needs to be run the first time - maybe an update script or something similar.
let conf = new Conf(); // instantiate without encryption key
let store = conf.store;
// Set the encryption key.
conf.encryptionKey = 'my_key';
// Conf just needs to perform a write then the written data will be successfully decrypted later on.
conf.store = store;
```

@sindresorhus do you think `Conf` should test the read data and handle this case, or should we just make a little info about it in the readme/wiki?

callodacity:
I suppose we should catch the error as well in this catch block so `Conf` will silently just reset the data to `{}` if the data can't be decrypted.

https://github.com/sindresorhus/conf/blob/52c3d3018ace7725df0a264165ca69d4ad624546/index.js#L115-L126

SamVerschueren:
I don't think resetting the data makes sense. Why not just return the data as is?

Another approach could be like we do in `cache-conf`. Set an object with a `data` property and an `encryption` flag.

```
{
   "foo" : {
       "encryption": true,
       "data": " abc"
    }
}
```

Or set the used algorithm as encryption value to make it more future proof. If `encryption` is missing, we could auto-upgrade if we detect an encryption key with unencrypted data.

callodacity:
Well the error caught would be because the data couldn't be decrypted - which could mean that it was either tampered with, or incorrectly encrypted, etc. Part of the reason to add this was:

> If the file is changed in any way, the decryption will not work, in which case the store will just reset back to its default state.

Of course, it could mean that the data wasn't even encrypted in the first place, and is just valid JSON (in this case I think it's fine to just return the data).

sindresorhus:
> The best approach IMO is something like to you said just to auto-upgrade if we detect an encryptionKey but we have unencrypted data. Probably keeping everything else as-is would keep things simple.

:+1:

PR #22, opened 2017-08-27
Add support for observing key paths

callodacity:
Fixes https://github.com/sindresorhus/electron-store/issues/14

Inspired partly by [electron-settings](https://github.com/nathanbuchar/electron-settings/wiki/API-documentation#watch).

@sindresorhus if you'd like anything else to be added, let me know 👍

### DIFF
--- a/index.js
+++ b/index.js
@@ -101,6 +101,7 @@ class Conf {
 		makeDir.sync(path.dirname(this.path));
 
 		let data = JSON.stringify(val, null, '\t');
+
 		if (this.encryptionKey) {
 			const cipher = crypto.createCipher('aes-256-cbc', this.encryptionKey);
 			data = Buffer.concat([cipher.update(Buffer.from(data)), cipher.final()]);
--- a/readme.md
+++ b/readme.md
@@ -80,16 +80,16 @@ The only use-case I can think of is having the config located in the app directo
 
 #### encryptionKey
 
-Type: `string | Buffer | TypedArray | DataView`<br>
+Type: `string` `Buffer` `TypedArray` `DataView`<br>
 Default: `undefined`
 
-Note that this **is not intended for security purposes**, since the encryption key would be easily found inside a plain-text NodeJS application.
+Note that this is **not intended for security purposes**, since the encryption key would be easily found inside a plain-text Node.js app.
 
-Its main use is for obscurity: if a user was looking through the user config directory and found the config file, since it's a simple json file they may be tempted to modify/add/remove a key/value pair. By providing an encryption key `Conf` will obfuscate the file which should hopefully deter any users from doing so.
+Its main use is for obscurity. If a user looks through the config directory and finds the config file, since it's just a JSON file, they may be tempted to modify it. By providing an encryption key, the file will be obfuscated, which should hopefully deter any users from doing so.
 
-It also has the added bonus that it will maintain a config file's integrity: if the file is changed in any way then decryption will not work, in which case the store will just reset back to its default state.
+It also has the added bonus of ensuring the config file's integrity. If the file is changed in any way, the decryption will not work, in which case the store will just reset back to its default state.
 
-When specified, the `Conf` module will encrypt the store using the `"aes-256-cbc"` encryption algorithm.
+When specified, the store will be encrypted using the [`aes-256-cbc`](https://en.wikipedia.org/wiki/Block_cipher_mode_of_operation) encryption algorithm.
 
 ### Instance
 


callodacity:
@sindresorhus is there anything else you'd like me to do before this PR can be merged?

index.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 const fs = require('fs');
 const path = require('path');
+const assert = require('assert');
+const EventEmitter = require('events').EventEmitter;
```
sindresorhus: This can just be:

```js
const EventEmitter = require('events');
```

index.js
```diff
@@ -12,8 +14,10 @@ const obj = () => Object.create(null);
 delete require.cache[__filename];
 const parentDir = path.dirname((module.parent && module.parent.filename) || '.');
 
-class Conf {
+class Conf extends EventEmitter {
```
sindresorhus: There's no point in extending EventEmitter if we're not going to use its methods.

sindresorhus:
I think we should have two methods, one that watches the key path and reports on any writes and one that only reports on changed values. `observe()` and `onDidChange()`. See: https://atom.io/docs/api/v1.0.0/Config#instance-onDidChange

index.js
callodacity: I had it emitting a `change` event when the store was updated (L118), but that's really only about it.
Otherwise we could potentially store all watch functions in a local array and call them from there.

callodacity: We could just simply hook into `config.set` and then call the `onDidChange` callbacks from there?

sindresorhus: Just init a new EventEmitter instance in the Conf init constructor and use it internally as a property.

callodacity: 👍 I'll do that.

callodacity:
Okay, so I'll split `watch` into `observe` and `onDidChange`?

sindresorhus:
`observe()` is unconditionally called for every `.set()`, while `onDidChange()` is called only if the new value is different from the old one.

callodacity:
Looks like [`observe()` is just called immediately](https://github.com/atom/atom/blob/master/src/config.coffee#L966) with the given value, and then it just does an `onDidChange()` as normal (which checks if the new value is differen

sindresorhus:
@callodacity How is that different from what I described?

sindresorhus:
We also have to consider what to do when the user does `.delete()`. I guess the event should be triggered, but what should the new value be? `undefined`? If so, it depends on https://github.com/sindresorhus/conf/issues/25.

callodacity:
> observe() is unconditionally called for every .set(), while onDidChange() is called only if the new value is different from the old one.

The source for Atom's `observe()` is essentially this:
```coffeescript
observeKeyPath: (keyPath, options, callback) ->
    callback(@get(keyPath))
    @onDidChangeKeyPath keyPath, (event) -> callback(event.newValue)
```

So if I understand it correctly:

```js
config.observe('my.key.path', (value) => console.log(value));
config.set('my.key.path', true); // "true" is logged
config.set('my.key.path', true); // "true" it NOT logged, since `onDidChange()` is only called when the new value is different from the old one
```

sindresorhus:
Alright, then I think we should only do `onDidChange()` (much clearer method name too). I don't really see the point in `observe()` then. All users care about are things that changed.

sindresorhus:
> I think the best bet is that the listener just be automatically removed in this case.

Isn't that too magic? The user could bring the value back and some point by setting it again. I would prefer the user to be explicit about unsubscribing.

callodacity:
> The user could bring the value back and some point by setting it again.

I hadn't actually thought about that case - I was only considering users who would call `delete()` and assume that everything would be cleaned up. But I think leaving it up to the user to unsubscribe would probably be the simplest and most user-friendly in the long run.

test.js
```diff
@@ -209,3 +209,28 @@ test('fallback to cwd if `module.filename` is `null`', t => {
...
+	unsubscribeFoo();
+	t.context.conf.set('foo', fixture);
+
+	t.context.conf.set('baz.boo', fixture);
+	const unsubscribeBaz = t.context.conf.onDidChange('baz.boo', check);
+	t.context.conf.set('baz.boo', '🐴');
+	unsubscribeBaz();
+	t.context.conf.set('baz.boo', fixture);
+
+	if (runs !== 2) {
```
sindresorhus: You can use `t.plan()` instead.

test.js
```diff
@@ -209,3 +209,28 @@ test('fallback to cwd if `module.filename` is `null`', t => {
 	module.filename = preservedFilename;
 	del.sync(conf.path, {force: true});
 });
+
+test('onDidChange()', t => {
+	let runs = 0;
```
sindresorhus:
```js
const conf = t.context.conf;
```

To reduce the verboseness.

test.js
```diff
@@ -209,3 +209,28 @@ test('fallback to cwd if `module.filename` is `null`', t => {
...
+	t.context.conf.set('baz.boo', fixture);
+	const unsubscribeBaz = t.context.conf.onDidChange('baz.boo', check);
+	t.context.conf.set('baz.boo', '🐴');
+	unsubscribeBaz();
+	t.context.conf.set('baz.boo', fixture);
+
+	if (runs !== 2) {
+		t.fail();
+	}
+});
```
sindresorhus: Can you add a test with using `.delete()` to ensure the newValue is correct then (`undefined`), and then using `.set()` to ensure the old value is `undefined`.

index.js
```diff
@@ -67,6 +70,26 @@ class Conf {
...
+				assert.deepEqual(newValue, oldValue);
+			} catch (err) {
+				currentValue = newValue;
+				callback.call(this, newValue, oldValue);
+			}
+		};
+
+		this.events.on('change', onChange);
+
+		// Return a function which removes the listener.
```
sindresorhus: This comment is also just describing what is clear in the code, IMHO.

callodacity: I agree with you about all these comments too, but I was just erring on the side of clarity 👍

callodacity: I'll make it clear in the readme that it returns a function to unsubscribe, anyway.

callodacity:
@sindresorhus Atom's `onDidChange` says:

    If `keyPath` is not specified, your callback will be called on changes to any key.

Did we want to have the same functionality in `Conf`?

sindresorhus:
> Did we want to have the same functionality in Conf?

Yes, I think that could be useful.

callodacity:
Okay I think this one should be good to merge.

That last Travis build didn't pass - seems to be an issue which the container not this module though.

readme.md
```diff
@@ -104,6 +104,10 @@ Check if an item exists.
 
 Delete an item.
 
+#### .onDidChange([key], callback)
+
+Watches the given `key`, calling `callback` on any changes. If `key` is not given, then `callback` will be called after any change is made to the store.
```
sindresorhus: We need to document what parameters `callback` provides. And that `oldValue` is `undefined` when a key is first set and `newValue` is `undefined` if a key is deleted.

What should the `newValue`/`oldValue` be when the user doesn't specify a `key`? I guess the store itself, but is that really useful? I'm trying to think of actual use-cases for this feature? What does Atom return in that case, for example.

test.js
```diff
@@ -210,3 +210,73 @@ test('fallback to cwd if `module.filename` is `null`', t => {
...
+	conf.set('foo', fixture);
+});
+
+test('onDidChange() without key', t => {
+	const conf = t.context.conf;
+	t.plan(4);
+
+	let unsubscribe;
+
+	unsubscribe = conf.onDidChange((newValue, oldValue) => {
```
sindresorhus: Just put the `let` here. Same with the other test.

readme.md
callodacity: [Atom returns the store itself](https://github.com/atom/key-path-helpers/blob/master/lib/key-path-helpers.js#L17). I personally don't think I'd ever use this to watch the whole store though.

SamVerschueren:
@callodacity can you rebase against master to resolve merge conflicts?

callodacity:
@SamVerschueren sure thing, I’ll do it as soon as I’m at my computer again 👍

sindresorhus:
@callodacity I think you did something weird. It now includes the encryption changes too.

sindresorhus:
I'm still a little bit uncertain whether we should include support for no key in `onDidChange`. I've thought a lot about it and I have yet to come up with a use-case for it. Which is usually a sign I shouldn't include it.

sindresorhus:
@SamVerschueren What do you think? I'm talking about:

> If no `key` is given, then `callback` will be called whenever a change is made to the store.

kevva:
> If no key is given, then callback will be called whenever a change is made to the store.

Might be valuable if you could get the `key` in the callback in case you want to listen on multiple but don't explicitly want to set a `onDidChange` for all of them. But that's just a matter of taste.

```js
conf.onDidChange('foo', doSomeThing);
conf.onDidChange('bar', doSomeThing);
conf.onDidChange('unicorn', doAnotherThing);

// vs

conf.onDidChange((val, key) => {
     if (key === 'foo' || key ==='bar') {
          doSomeThing(val);
    }

     doAnotherThing(val);
});
```

However, I'm failing to see a use case for calling it every time for what you're describing so I'm leaning towards removing it.

sindresorhus:
@kevva Note that we currently just return the whole store when no key is specified, so your example is slightly more useful than reality, but I too still don't see much use-case.

callodacity:
I'm sorry guys, I messed this branch up quite nicely - I'll just create a new PR

SamVerschueren:
Because the main discussion is in here, let me answer it here :).

> If no key is given, then callback will be called whenever a change is made to the store.

I think we should enforce passing in a key and do not allow no key. As you said, we can't come up with a decent use case. We can always improve it later if someone comes up with a decent explanation as to why it could be a good addition.

callodacity:
I think that's a good move - if the user wants the changed store they can just use `conf.store` anyway.

PR #26, opened 2017-09-11
Add onDidChange() method

callodacity:
This is a new PR to fix the mess I made in #22 -- sorry about that!

I did make some changes, which we can discuss:

- Added type checking to arguments
- Removed support for listening to changes without keys
- ~~Added `key` as a third argument to the callback~~

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,6 +77,7 @@ class Conf {
 		if (typeof key !== 'string') {
 			throw new TypeError(`Expected \`key\` to be of type \`string\`, got ${typeof key}`);
 		}
+
 		if (typeof callback !== 'function') {
 			throw new TypeError(`Expected \`callback\` to be of type \`function\`, got ${typeof callback}`);
 		}
--- a/index.js
+++ b/index.js
@@ -76,21 +76,22 @@ class Conf {
 	onDidChange(key, callback) {
 		if (typeof key !== 'string') {
 			throw new TypeError(`Expected \`key\` to be of type \`string\`, got ${typeof key}`);
-		} else if (typeof callback !== 'function') {
+		}
+		if (typeof callback !== 'function') {
 			throw new TypeError(`Expected \`callback\` to be of type \`function\`, got ${typeof callback}`);
 		}
 
-		let currentValue = key ? this.get(key) : this.store;
+		let currentValue = this.get(key);
 
 		const onChange = () => {
 			const oldValue = currentValue;
-			const newValue = key ? this.get(key) : this.store;
+			const newValue = this.get(key);
 
 			try {
 				assert.deepEqual(newValue, oldValue);
 			} catch (err) {
 				currentValue = newValue;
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -70,6 +73,30 @@ class Conf {
...
+
+		const onChange = () => {
+			const oldValue = currentValue;
+			const newValue = key ? this.get(key) : this.store;
+
+			try {
+				assert.deepEqual(newValue, oldValue);
+			} catch (err) {
+				currentValue = newValue;
+				callback.call(this, newValue, oldValue, key);
```
kevva: It doesn't make sense to have `key` as the third argument when the user already knows about it, IMO.

index.js
```diff
@@ -70,6 +73,30 @@ class Conf {
 	clear() {
 		this.store = obj();
 	}
+	onDidChange(key, callback) {
+		if (typeof key !== 'string') {
+			throw new TypeError(`Expected \`key\` to be of type \`string\`, got ${typeof key}`);
+		} else if (typeof callback !== 'function') {
```
kevva: Don't need `else if` here. Just do a new `if` block.

index.js
```diff
@@ -70,6 +73,30 @@ class Conf {
...
+			throw new TypeError(`Expected \`callback\` to be of type \`function\`, got ${typeof callback}`);
+		}
+
+		let currentValue = key ? this.get(key) : this.store;
```
kevva: I don't think it should fall back to `this.store`. Just do `let currentValue = this.get(key);` and the same with `newValue` further down.

SamVerschueren:
This is a little off topic and is more a discussion on the API of this feature, not really regarding the PR. Did we consider using Observables for this use case? 

```js
const Conf = require('conf');

const config = new Conf();

config.select('unicorn').subscribe(value => {
    console.log(value);
    //=> 🌈
});

config.set('unicorn', '🌈');
```

The benefit is that you can easily integrate it together with other observables or use other operators to manipulate the result before returning.

The `select` method is based upon how [ngrx](https://github.com/ngrx/platform) (Redux for Angular) does it. We can always use another name off course.

It's just an idea though. Want to make sure we considered all our options :).

SamVerschueren:
Another approach could be like we did with [cache-conf](https://github.com/samverschueren/cache-conf). We keep `conf` small and focused, and this behaviour could be put in an `observable-conf` package which is a drop-in replacement for `conf`.

SamVerschueren:
I discussed this with @sindresorhus and decided that this feature should be in the core. However, it's nice if we have a wrapper for observables, so I created [observable-conf](https://github.com/SamVerschueren/observable-conf). I can use this `onDidChange()` feature when it lands.

callodacity:
@SamVerschueren no worries, that sounds good to me. Do you know if we are we waiting for anything before merging this into core?

sindresorhus:
@callodacity Was just waiting on me having time to merge it. Thanks for working on this 😀

PR #27, opened 2017-09-17
Check if data is unencrypted before encrypting

callodacity:
For the issue raised in #21.

I think this solution is simple enough:

- Try to parse and return the data. If not, try and decrypt it.
- If decrypting fails, act the same as before (as if the file didn't exist, or there was a `SyntaxError`).
- Otherwise, if parsing worked - just continue as normal.

So, notable changes:

1. Users can now set `encryptionKey` without worrying about upgrading the store
2. If decryption fails, `Conf` will silently just return the defaults

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,8 +107,12 @@ class Conf {
 			let data = fs.readFileSync(this.path, this.encryptionKey ? null : 'utf8');
 
 			if (this.encryptionKey) {
-				const decipher = crypto.createDecipher('aes-256-cbc', this.encryptionKey);
-				data = Buffer.concat([decipher.update(data), decipher.final()]);
+				try {
+					return Object.assign(obj(), JSON.parse(data));
+				} catch (err) {
+					const decipher = crypto.createDecipher('aes-256-cbc', this.encryptionKey);
+					data = Buffer.concat([decipher.update(data), decipher.final()]);
+				}
 			}
 
 			return Object.assign(obj(), JSON.parse(data));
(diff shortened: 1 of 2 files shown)


callodacity:
If a user *has had* encryption enabled, and decides to unset it, then their data will be lost and it will return to defaults. I don't see a great way around this since we can't know how to decrypt the data if we don't have an encryption key. I think just returning the defaults is a good approach.

In any case, the user can remove encryption without data loss like so:

```js
let conf = new Conf({ encryptionKey: 'my_key' });
conf.set('foo', 'bar');
const store = conf.store;

conf = new Conf();  // no key
conf.store = store; // restore the data
```

index.js
```diff
@@ -107,8 +107,16 @@ class Conf {
 			let data = fs.readFileSync(this.path, this.encryptionKey ? null : 'utf8');
 
 			if (this.encryptionKey) {
-				const decipher = crypto.createDecipher('aes-256-cbc', this.encryptionKey);
-				data = Buffer.concat([decipher.update(data), decipher.final()]);
+				try {
+					return Object.assign(obj(), JSON.parse(data));
```
sindresorhus: Wouldn't it make more sense to try to decrypt first and rather fall back to `Object.assign(obj(), JSON.parse(data));` if decryption fails? Your way here adds a slight overhead each time instead of just the first.

sindresorhus:
> I think just returning the defaults is a good approach.

Agreed. When https://github.com/sindresorhus/electron-store/issues/19 is implemented, we can make sure it's possible to move off or to another encryption method/key in a migration.

index.js
callodacity: Good point, I think I was over-thinking the whole thing actually, since that was my first approach!

PR #36, opened 2018-03-25
fixes #35

### LINKED ISSUE #35: Make the -nodejs suffix configurable
Related: https://github.com/sindresorhus/env-paths/issues/2

MegaGM:


sindresorhus:
This PR needs a proper title.

sindresorhus:
@MegaGM Would you be able to document it? Also, include your use-case about wanting to use to the same store for both Node.js and Electron, so it's clear to users what they would use the option for.

Would also be great if you could include a test. You could do something similar to https://github.com/sindresorhus/conf/blob/d22f8237241c2b7dc98ef52065a3e2aa7043c380/test.js#L158 by just checking that the `.path` property include your chosen suffix.

sindresorhus:
Friendly ping :)

PR #38, opened 2018-04-08
Avoiding always writing to disk upon instantiation

### LINKED ISSUE #32: Problematic file write upon instantiation
I have a bitbar plugin that gets refreshed every second, 99% of the times it just reads data from the configuration without mutating it in any way.

In `conf`'s constructor there's this line:
```
this.store = Object.assign(obj(), opts.defaults, this.store);
```

Which calls `store`'s setter method, which writes back to disk the value of the store we just read. Basically every instantiation causes a write, this may be problematic in some use cases.

This is a problem for me because I have the actual `json` file stored in the dropbox folder, and being it constantly re-written every second it causes Dropbox to consume 20% of my CPU just for syncing this one file that actually almost never changes.

Is this behavior intentional? If so, would you consider a PR adding an option for disabling it?

fabiospampinato:
Fixes #32.

A couple of things worth noting:

- that `assert.deepEqual` call may be expensive, but hopefully less expensive than a file write.
- if a `new Conf` is instantiated, without any default properties, `conf.path` won't exist until `conf.set` is called.

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,7 +41,14 @@ class Conf {
 		this.events = new EventEmitter();
 		this.encryptionKey = options.encryptionKey;
 		this.path = path.resolve(options.cwd, `${options.configName}.json`);
-		this.store = Object.assign(plainObject(), options.defaults, this.store);
+
+		const fileStore = this.store;
+		const store = Object.assign(plainObject(), options.defaults, this.store);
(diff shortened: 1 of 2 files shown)


sindresorhus:
Looks good. Thanks :)

PR #39, opened 2018-04-28
Amend issues with babelify

EdenCoder:
This does not function with babelify as `require.cache` is undefined, it works fine without deleting the cache in babelify


sindresorhus:
Thanks, but this should be fixed in Babelify. `require.cache` is expected to exist in Node.js, and it's up to Babelify to be compatible. Every module should not have to work around a single tool. Open an issue on Babelify instead ;)

PR #40, opened 2018-04-29
more info on default values

burningTyger:
I had to test this manually to see what happens with default values. Might as well add this to the Readme.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -57,6 +57,8 @@ Type: `Object`
 
 Default config.
 
+If `key` already exists in config file the value doesn't change. If the key or the config file don't yet exist they will be created and set to the default value.
+
 #### configName
 
 Type: `string`<br>


readme.md
```diff
@@ -57,6 +57,8 @@ Type: `Object`
 
 Default config.
 
+If `key` already exists in config file the value doesn't change. If the key or the config file don't yet exist they will be created and set to the default value.
```
sindresorhus: I don't mind documenting this, but I think it could be written more succinctly.

burningTyger: I really don't mind if you have a better sentence in mind. It thought writing a pull request is as good as writing an issue :) It does sound a bit awkward...

burningTyger: How about "Creates and sets the default values for `key` and writes them to disk unless they already exist."
BTW you have an awesome amount of cool repos. Just noticed :)

PR #42, opened 2018-05-30
Add option to validate JSON file

bampakoa:
If the option `validate` is true and the configuration file is not a valid JSON, it throws an exception. References https://github.com/sindresorhus/electron-store/issues/35


index.js
```diff
@@ -43,6 +43,14 @@ class Conf {
 		this.encryptionKey = options.encryptionKey;
 		this.path = path.resolve(options.cwd, `${options.configName}.${options.fileExtension}`);
 
+		if (options.validate) {
+			try {
+				JSON.parse(fs.readFileSync(this.path));
+			} catch (ex) {
+				throw ex;
+			}
```
sindresorhus: I don't understand why you're wrapping this in try/catch? It has no purpose.

And I think this check would be better done here instead: https://github.com/sindresorhus/conf/blob/8ce463d1ecda92b53595904e01868012c5d11659/index.js#L152

sindresorhus:
Sorry about the slow reply. The option needs to be documented too.

index.js
bampakoa: I added the try/catch in order to throw the exception to the consumer of the library so that he can handle it appropriately.

I added the check in the constructor because `options` object is available there and I thought that it would be a good choice to validate early on.

bampakoa:
@sindresorhus this is ready

sindresorhus:
Thinking about this a bit more, I think it would be better to handle this in https://github.com/sindresorhus/conf/pull/42/files#diff-168726dbe96b3ce427e7fedce31bb0bcR148 Then you also don't have to read the JSON twice.

bampakoa:
@sindresorhus I set the validate option as:

sindresorhus:
Hard to say without seeing your code. Can you push what you got?

bampakoa:
sure, here it is

test.js
```diff
@@ -331,3 +331,9 @@ test('doesn\'t write to disk upon instanciation if and only if the store didn\'t
 	exists = fs.existsSync(conf.path);
 	t.is(exists, true);
 });
+
+test('`validate` option', t => {
+	fs.writeFileSync(t.context.conf.path, 'Some text');
```
sindresorhus: You cannot use the auto-generated instance here. The `.path` is different for each instance.

index.js
```diff
@@ -44,7 +44,9 @@ class Conf {
 		this.encryptionKey = options.encryptionKey;
 
 		const fileExtension = options.fileExtension ? `.${options.fileExtension}` : '';
-		this.path = path.resolve(options.cwd, `${options.configName}${fileExtension}`);
+		this.path = path.resolve(options.cwd, `${options.configName}.${options.fileExtension}`);
+    
+		this.validate = options.validate;
```
```suggestion
		this._validate = options.validate;
```

It's internal and should be marked as such.

index.js
```diff
@@ -138,6 +140,10 @@ class Conf {
 				} catch (_) {}
 			}
 
+			if (this.validate) {
+				JSON.parse(data);
+			}
```
sindresorhus: This is incorrect. It will never pass as it's just the same as the parse on line 147.

bampakoa:
here they are @sindresorhus. Thanks for your suggestions! I also added a second test for valid data.

test.js
```diff
@@ -331,3 +331,15 @@ test('doesn\'t write to disk upon instanciation if and only if the store didn\'t
 	exists = fs.existsSync(conf.path);
 	t.is(exists, true);
 });
+
+test('`validate` option with invalid data', t => {
+	const conf = new Conf({cwd: tempy.directory(), validate: true});
+	fs.writeFileSync(conf.path, 'Non-JSON data');
+	t.deepEqual(conf.store, {});
```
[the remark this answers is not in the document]
bampakoa: It will never throw because of this check in line 155:

```
if (error.name === 'SyntaxError') {
  return plainObject();
}
```

`JSON.parse` throws `SyntaxError` which is already caught. We can either try/catch the `JSON.parse` in line 143-146 and throw error there, which you have suggested not to do, or remove the check for `SyntaxError`.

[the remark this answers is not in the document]
bampakoa: @sindresorhus what do you think about that?

bampakoa:
@sindresorhus in the last commit, I throw an error with custom message when validate is enabled to bypass the caught `SyntaxError` error later on.

sindresorhus:
The implementation is still suboptimal and I unfortunately don't have more time to guide you through this.

PR #43, opened 2018-06-21
Make the config file private to its owner

simon-engledew:
The current permissions on the config file allows other users on the system to read the data (0o644).

It's worth throwing away the read permissions for group and public.

### DIFF
--- a/index.js
+++ b/index.js
@@ -161,7 +161,7 @@ class Conf {
 			data = Buffer.concat([cipher.update(Buffer.from(data)), cipher.final()]);
 		}
 
-		writeFileAtomic.sync(this.path, data);
+		writeFileAtomic.sync(this.path, data, {mode: 0o600});
 		this.events.emit('change');
 	}
 


sindresorhus:
I intentionally didn't do this here as it caused a myriad of problems when I did it with a previous similar module called [`configstore`](https://github.com/yeoman/configstore): https://github.com/search?q=EACCES%3A+permission+denied+configstore&type=Issues Basically people ran tools using configstore with `sudo` and then later they did not and suddenly they couldn't read the config file.

sindresorhus:
I also don't really see the point as the config is already saved somewhere inside the user's home directory, which shouldn't be accessible to other accounts, and if it is, they have access to everything else anyway.

simon-engledew:
Not really sure what to say to that!

People running or installing NPM packages with `sudo` is very scary. I am sad that happens enough that you have had to design a library around it. :)

The real user is available in $SUDO_USER if you wanted to ensure the correct owner of the directory. `chown($SUDO_USER or $USER)`.

I think home directories are usually world accessible by default, although God knows why. I just booted up a Linux VM to check and I can happily read files that belong to another user in their home directory. :/ I thought you could still read files inside directories with no `go+x`, but it turns out my mental model was incorrect. Still, in any directory with `go+x` files are fair game if the read flag is set.  On macOS `conf` puts the file in the ~/Library directory which is indeed locked down. If this is true for every other platform I guess the PR is moot. :)

PR #47, opened 2018-09-06
Trigger `onDidChange` when the file is altered

npezza93:
Fixes sindresorhus/electron-store#39

### DIFF
--- a/index.js
+++ b/index.js
@@ -213,10 +213,9 @@ module.exports = class Conf {
 		if (!fs.existsSync(this.path)) {
 			writeFileAtomic.sync(this.path, JSON.stringify({}, null, '\t'));
 		}
-		chokidar.watch(this.path).on('all', event => {
-			console.log(event);
-			this.events.emit('change');
-		});
+		chokidar.watch(this.path)
+			.on('add', () => this.events.emit('change'))
+			.on('change', () => this.events.emit('change'));
 	}
 
 	// TODO: Use `Object.entries()` when targeting Node.js 8
--- a/index.js
+++ b/index.js
@@ -213,8 +213,9 @@ module.exports = class Conf {
 		if (!fs.existsSync(this.path)) {
 			writeFileAtomic.sync(this.path, JSON.stringify({}, null, '\t'));
 		}
-
+		console.log('watching');
 		chokidar.watch(this.path).on('change', () => {
+			console.log('chokidar changed');
 			this.events.emit('change');
 		});
 	}
--- a/index.js
+++ b/index.js
@@ -208,7 +208,7 @@ module.exports = class Conf {
 		this._ensureDirectory();
 
 		let wait = false;
-		fs.watch(path.dirname(this.path), {encoding: this.encryptionKey ? null : 'utf8'}, () => {
+		fs.watch(this.path, () => {
 			if (wait) {
 				return;
 			}
--- a/index.js
+++ b/index.js
@@ -213,9 +213,8 @@ module.exports = class Conf {
 		if (!fs.existsSync(this.path)) {
 			writeFileAtomic.sync(this.path, JSON.stringify({}, null, '\t'));
 		}
-		console.log('watching');
-		chokidar.watch(this.path).on('change', () => {
-			console.log('chokidar changed');
+		chokidar.watch(this.path).on('all', event => {
+			console.log(event);
 			this.events.emit('change');
 		});
 	}
--- a/index.js
+++ b/index.js
@@ -77,7 +77,7 @@ module.exports = class Conf {
 		} catch (_) {
 			this.store = store;
 		}
-		this.watch();
+		this._watch();
 	}
 
 	get(key, defaultValue) {
@@ -204,7 +204,7 @@ module.exports = class Conf {
 		makeDir.sync(path.dirname(this.path));
 	}
 
-	watch() {
+	_watch() {
 		this._ensureDirectory();
 
 		let wait = false;
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const makeDir = require('make-dir');
 const pkgUp = require('pkg-up');
 const envPaths = require('env-paths');
 const writeFileAtomic = require('write-file-atomic');
+const debounceFn = require('debounce-fn');
 
 const plainObject = () => Object.create(null);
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -165,6 +165,26 @@ class Conf {
 		this.events.emit('change');
 	}
 
+	createDir() {
```
sindresorhus: `createDir` => `_ensureDirectory`

index.js
```diff
@@ -165,6 +165,26 @@ class Conf {
...
+		// Ensure the directory exists as it could have been deleted in the meantime
+		makeDir.sync(path.dirname(this.path));
+	}
+
+	watch() {
```
sindresorhus: `_watch() {`

index.js
```diff
@@ -165,6 +165,26 @@ class Conf {
...
+		this.createDir();
+
+		let wait = false;
+		fs.watch(path.dirname(this.path), {encoding: this.encryptionKey ? null : 'utf8'}, () => {
```
sindresorhus: `fs.watch` accepts a filename, so no need to get the directory.

The `encoding` option setting here is incorrect.

index.js
```diff
@@ -165,6 +165,26 @@ class Conf {
...
+			if (wait) {
+				return;
+			}
+			wait = setTimeout(() => {
+				wait = false;
+			}, 100);
+			this.events.emit('change');
```
sindresorhus: You have to check `eventType` for the correct events.

index.js
```diff
@@ -165,6 +165,26 @@ class Conf {
...
+		this.createDir();
+
+		let wait = false;
+		fs.watch(path.dirname(this.path), {encoding: this.encryptionKey ? null : 'utf8'}, () => {
+			if (wait) {
+				return;
+			}
+			wait = setTimeout(() => {
+				wait = false;
+			}, 100);
```
sindresorhus: Use https://github.com/sindresorhus/debounce-fn

sindresorhus:
This needs documentation updates and a test.

index.js
```diff
@@ -168,6 +171,22 @@ class Conf {
 		this.events.emit('change');
 	}
 
+	_ensureDirectory() {
+		// Ensure the directory exists as it could have been deleted in the meantime
+		makeDir.sync(path.dirname(this.path));
+	}
+
+	_watch() {
```
npezza93: I'm leaning towards making this a public method, that way we can return the watcher/a disposable that way the watcher can get cleaned up. Curious what your opinion on this is.

sindresorhus:
Can you revert https://github.com/sindresorhus/conf/pull/47/commits/70e619f8c75fa59f460702bdd3be8748cef48565? `chokidar` is a huge dependenyc and not worth it for the tiny use-case here.

index.js
sindresorhus: I would prefer to wait until someone actually requests that.

sindresorhus:
@npezza93 Bump :)

sindresorhus:
Can you also fix the merge conflict?

readme.md
```diff
@@ -145,6 +145,16 @@ You can pass an empty string to remove the suffix.
 
 For example, on macOS, the config file will be stored in the `~/Library/Preferences/foo-nodejs` directory, where `foo` is the `projectName`.
 
+#### watch
+
+type: `Boolean`<br>
+Default: `false`
+
+Sets a file watcher.
+
```
sindresorhus: The docs here needs to be improved. `Sets a file watcher.` is not a good or useful description. It also has no mention of triggering `onDidChange`.

readme.md
```diff
@@ -145,6 +145,16 @@ You can pass an empty string to remove the suffix.
...
 For example, on macOS, the config file will be stored in the `~/Library/Preferences/foo-nodejs` directory, where `foo` is the `projectName`.
 
+#### watch
+
+type: `boolean`<br>
+Default: `false`
+
+Sets a file watcher.
+
+This is useful if you have more than one instance reading the config. If one instance changes a config, this allows it to get propagated to all other instances. If you only ever have one instance in your app you can keep this off.
```
sindresorhus: It's not really just about instances, as I cannot see why anyone would have multiple instances in the same process. This is about getting updates between different processes.

test.js
```diff
@@ -434,3 +434,25 @@ test('`clearInvalidConfig` option - valid data', t => {
...
+	const checkFoo = (newValue, oldValue) => {
+		t.is(newValue, '🐴');
+		t.is(oldValue, '👾');
+		conf2.set('foo', '👾');
+		t.end();
+	};
+
+	t.is(conf2.get('foo'), '👾');
+	t.is(conf1.path, conf2.path);
+	conf1.onDidChange('foo', checkFoo);
```
sindresorhus: You can use https://github.com/sindresorhus/p-event here to be able to use `test` and async/await instead of `test.cb`.

index.js
```diff
@@ -199,6 +202,26 @@ module.exports = class Conf {
...
+
+		if (!fs.existsSync(this.path)) {
+			writeFileAtomic.sync(this.path, JSON.stringify({}, null, '\t'));
+		}
+
+		fs.watch(this.path, debounceFn(eventType => {
+			if (eventType === 'change') {
+				this.events.emit('change');
+			}
+		}, {wait: 100}));
```
sindresorhus: You also need the `persistence: false` option. Please read the `fs.watch` docs more closely.

index.js
```diff
@@ -199,6 +202,26 @@ module.exports = class Conf {
...
+		// TODO: Use `fs.mkdirSync` `recursive` option when targeting Node.js 12
+		makeDir.sync(path.dirname(this.path));
+	}
+
+	_watch() {
+		this._ensureDirectory();
+
+		if (!fs.existsSync(this.path)) {
+			writeFileAtomic.sync(this.path, JSON.stringify({}, null, '\t'));
+		}
```
sindresorhus: This will not work as you're not handling other writing options like when it's encrypted. You should instead reuse the existing writing logic.

sindresorhus:
Bump

zeke:
Hey @npezza93, thanks for working on this. My app would definitely benefit from it. 🤞

sindresorhus:
Closing this in favor of https://github.com/sindresorhus/conf/issues/75.

PR #49, opened 2018-10-13
Add store migrations

npezza93:


index.js
```diff
@@ -50,6 +52,8 @@ class Conf {
 		} catch (e) {
 			this.store = store;
 		}
+
+		this.migrate(options, options.projectVersion);
```
sindresorhus: I think you should do the if conditional here instead of inside `migrate()`.

index.js
```diff
@@ -92,6 +96,26 @@ class Conf {
 		this.store = plainObject();
 	}
 
+	migrate(options, expectedVersion) {
```
sindresorhus: It should be marked internal.

```suggestion
	_migrate(options, expectedVersion) {
```

index.js
sindresorhus: Passing in `options.projectVersion` seems moot as you're already passing in the `options` object.

test.js
```diff
@@ -321,3 +321,52 @@ test('doesn\'t write to disk upon instanciation if and only if the store didn\'t
...
+
+	t.not(conf.get('packageVersion'), undefined);
+	t.is(conf.get('packageVersion'), require('./package.json').version);
+});
+
+test('packageVersion is not set without migrations', t => {
+	const {conf} = t.context;
+
+	t.is(conf.get('packageVersion'), undefined);
+});
```
sindresorhus: There should be a test for what happens when an older version is used with newer config.

index.js
```diff
@@ -92,6 +96,26 @@ class Conf {
...
+					return semver.lte(version, expectedVersion) && semver.gt(version, runningVersion);
+				}).sort(semver);
+
+				for (const version of migrationsToRun) {
+					options.migrations[version](this);
+				}
+			}
+
+			if (runningVersion !== expectedVersion) {
+				this.set('packageVersion', expectedVersion);
```
sindresorhus: You can't just set a key. You could potentially silently be overwriting a user's key by the same name. The key must me something that is very unlikely to conflict. Maybe `__packageVersion__`

index.js
```diff
@@ -92,6 +96,26 @@ class Conf {
 		this.store = plainObject();
 	}
 
+	migrate(options, expectedVersion) {
+		if (options.migrations) {
+			const runningVersion = this.store.packageVersion || '0.0.0';
```
sindresorhus: This should be validated like the `packageName` is on line 31.

readme.md
```diff
@@ -106,6 +113,35 @@ Extension of the config file.
 
 You would usually not need this, but could be useful if you want to interact with a file with a custom file extension that can be associated with your app. These might be simple save/export/preference files that are intended to be shareable or saved outside of the app.
 
+#### migrations
+
+ type: `object`<br>
```
sindresorhus:
```suggestion
 type: `Object`<br>
```

test.js
```diff
@@ -321,3 +321,52 @@ test('doesn\'t write to disk upon instanciation if and only if the store didn\'t
...
+	t.is(conf2.get('4'), undefined);
+	t.is(conf2.get('new'), 'old');
+	t.is(conf2.get('2'), 1);
+	t.is(conf2.get('packageVersion'), require('./package.json').version);
+});
+
+test('packageVersion is set initially', t => {
+	const conf = new Conf({cwd: tempy.directory(), migrations: {}});
+
+	t.not(conf.get('packageVersion'), undefined);
```
sindresorhus: This assertion is moot. It's better tested below.

readme.md
```diff
@@ -106,6 +113,35 @@ Extension of the config file.
...
+ Default: `undefined`
+
+ Migrations to be run between versions.
+
+ Useful for transitioning conf changes between application versions. Ex:
+ ```js
+ // version is set to 0.0.1, the current app version is 2.0.8, old is set to 1
+ const store = new Conf({
+ 	migrations: {
+ 		'0.0.0': store => {
```
sindresorhus: This is a bad example, as there are no versions before this, it's not actually a migration.

index.js
```diff
@@ -92,6 +96,26 @@ class Conf {
...
+
+			if (semver.lt(runningVersion, expectedVersion)) {
+				const migrationsToRun = Object.keys(options.migrations).filter(version => {
+					return semver.lte(version, expectedVersion) && semver.gt(version, runningVersion);
+				}).sort(semver);
```
sindresorhus: Shouldn't this be `.sort(semver.compare)`?

readme.md
```diff
@@ -107,6 +114,35 @@ Extension of the config file.
...
+ 			store.set('new', old);
+ 			store.delete('old');
+ 		},
+ 		'1.0.2': store => {
+ 			store.set('a new key', 't');
+ 		}
+ 	}
+ });
+ console.log(store.store);
+ // { version: '2.0.8', new: 1, 'a new key': 't' }
```
sindresorhus: This is outdated

readme.md
```diff
@@ -107,6 +114,35 @@ Extension of the config file.
...
+ 			store.set('bad key', 2);
+ 		},
+ 		'1.0.0': store => {
+ 			const old = store.get('old');
+ 			store.set('new', old);
+ 			store.delete('old');
+ 		},
+ 		'1.0.2': store => {
+ 			store.set('a new key', 't');
+ 		}
```
sindresorhus: I think it could be useful to have a special `*` migration that is run for every version. If you need more advanced handling.

sindresorhus:
@npezza93 Still interested in finishing this? :)

sindresorhus:
Bump

readme.md
npezza93: Think I might leave that as a separate PR if that's cool.

index.js
```diff
@@ -35,22 +36,25 @@ const checkValueType = (key, value) => {
...
 			...options
 		};
 
 		if (!options.cwd) {
 			if (!options.projectName) {
-				const pkgPath = pkgUp.sync(parentDir);
 				// Can't use `require` because of Webpack being annoying:
 				// https://github.com/webpack/webpack/issues/196
-				options.projectName = pkgPath && JSON.parse(fs.readFileSync(pkgPath, 'utf8')).name;
+				options.projectName = pkg && pkg.name;
```
Hongarc: You should move comment too, not only code

sindresorhus:
@yaodingyd @Hongarc Would you be able to help review?

index.js
```diff
@@ -36,24 +36,25 @@ const checkValueType = (key, value) => {
...
 			clearInvalidConfig: true,
 			serialize: value => JSON.stringify(value, null, '\t'),
 			deserialize: JSON.parse,
+			projectVersion: pkg && pkg.version,
 			...options
 		};
 
 		if (!options.cwd) {
 			if (!options.projectName) {
-				const pkgPath = pkgUp.sync(parentDir);
```
sindresorhus: Moving `pkgPath` here was an intentional change. See: https://github.com/sindresorhus/conf/commit/bb26c4f0a64cd8f1868eda2883cea651f0462fc0

index.js
```diff
@@ -157,6 +165,27 @@ class Conf {
 		this.store = plainObject();
 	}
 
+	_migrate(options) {
+		const runningVersion = this.store.__packageVersion__ || '0.0.0';
```
sindresorhus: You need to ensure the user doesn't accidentally set/change `__packageVersion__`, which cause weird problems.

sindresorhus: `__packageVersion__` should also not be available in `this.store`. It's an internal value.

sindresorhus:
Instead of `__packageVersion__`, I think we should go with `__internal__`, so it's more flexible for possible future internal values. So it will be `__internal__.packageVersion`.

sindresorhus:
I think this is going to need more tests. It's a difficult feature to get right and I'm not confident in the code without more tests.

sindresorhus:
Closing this in favor of #74. Probably better for someone else to tackle this.

PR #50, opened 2018-10-13
Use isDeepStrictEqual instead of assert.deepEqual

npezza93:
Point to using Node >= 10 and fix some lints that resulted from that upgrade

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,6 +42,10 @@ module.exports = class Conf {
 			...options
 		};
 
+		if (!options.projectName && !options.cwd) {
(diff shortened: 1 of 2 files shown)


sindresorhus:
Can't really target Node.js 10 yet, unfortunately. The convention in the community is to follow Node.js LTS releases, so we're stuck targeting Node.js 6 for now. But I'll keep this open until we can switch.

npezza93:
That sounds great, thanks @sindresorhus!

PR #52, opened 2018-11-09
Add `projectSuffix` option

### LINKED ISSUE #35: Make the -nodejs suffix configurable

niktekusho:
Since #35 is still open, I thought I could finally resolve it maintaining the current behavior (as of 2.0.0) of adding the 'nodejs' suffix by default.

The option for controlling it is called `projectSuffix`, but feel free to change it to something you think is more appropriate.

I added 3 test cases which should cover all the cases I thought of:

1. **compatibility**: maintain the 2.0.0 behavior when not specifying the `projectSuffix`, so it will add the 'nodejs' suffix;
2. **new**: specifying an empty string should disable the suffix;
3. **new**: specifying a custom suffix should just use it instead of 'nodejs'.

I also updated the README with the new option specifics (added an `Examples` subsection).

---

Fixes #35

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,7 +34,6 @@ class Conf {
 		options = Object.assign({
 			configName: 'config',
 			fileExtension: 'json',
-			// Maintain behaviour of 2.0.0 for the new option
 			projectSuffix: 'nodejs'
 		}, options);
 
--- a/index.js
+++ b/index.js
@@ -33,11 +33,13 @@ class Conf {
 
 		options = Object.assign({
 			configName: 'config',
-			fileExtension: 'json'
+			fileExtension: 'json',
+			// Maintain behaviour of 2.0.0 for the new option
+			projectSuffix: 'nodejs'
 		}, options);
 
 		if (!options.cwd) {
-			options.cwd = envPaths(options.projectName).config;
+			options.cwd = envPaths(options.projectName, {suffix: options.projectSuffix}).config;
 		}
 
 		this.events = new EventEmitter();
--- a/readme.md
+++ b/readme.md
@@ -74,6 +74,63 @@ Default: The `name` field in the package.json closest to where `conf` is importe
 
 You only need to specify this if you don't have a package.json file in your project.
 
+#### projectSuffix
+
+Type: `string`<br>
+Default: `nodejs`
+
+Suffix to be added to `projectName` during config file creation.
+
+You can pass an empty string in order to disable any suffixes at all.
+
+##### Examples
+
+Every example in this subsection is supposedly used in a project called `test-project`.
+
+**Default option**
+
+```js
+const conf = new Conf();
+```
+
+The config file will be placed in the following subtree:
+```
+$OS_SPECIFIC_PATH
+├─── test-project-nodejs
+|   └─── Config
+|       └─── config.json
+```
+
+**Empty string option**
+
+```js
+const projectSuffix = '';
+const conf = new Conf({projectSuffix});
+```
+
+The config file will be placed in the following subtree:
+```
+$OS_SPECIFIC_PATH
+├─── test-project // <--- Here! No '-nodejs' suffix!
+|   └─── Config
+|       └─── config.json
+```
+
+**Custom string option**
+
+```js
+const projectSuffix = 'custom';
+const conf = new Conf({projectSuffix});
+```
+
+The config file will be placed in the following subtree:
+```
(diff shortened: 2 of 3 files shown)


readme.md
```diff
@@ -74,6 +74,63 @@ Default: The `name` field in the package.json closest to where `conf` is importe
...
 You only need to specify this if you don't have a package.json file in your project.
 
+#### projectSuffix
+
+Type: `string`<br>
+Default: `nodejs`
+
+Suffix to be added to `projectName` during config file creation.
+
+You can pass an empty string in order to disable any suffixes at all.
```
sindresorhus: This should explain why NOT specifying this option is good. From: https://github.com/sindresorhus/env-paths#suffix

readme.md
```diff
@@ -74,6 +74,63 @@ Default: The `name` field in the package.json closest to where `conf` is importe
...
+const conf = new Conf({projectSuffix});
+```
+
+The config file will be placed in the following subtree:
+```
+$OS_SPECIFIC_PATH
+├─── test-project-custom // <--- Here! We have our very own '-custom' suffix!
+|   └─── Config
+|       └─── config.json
+```
```
sindresorhus: While explaining things clearly in docs is a good thing, I think this is taking too far. One example is more than enough for this option.

index.js
```diff
@@ -33,11 +33,13 @@ class Conf {
 
 		options = Object.assign({
 			configName: 'config',
-			fileExtension: 'json'
+			fileExtension: 'json',
+			// Maintain behaviour of 2.0.0 for the new option
```
sindresorhus: Remove this comment. It doesn't add anything. It's clear from the context what it does.

readme.md
```diff
@@ -74,6 +74,63 @@ Default: The `name` field in the package.json closest to where `conf` is importe
...
+
+```js
+const conf = new Conf();
+```
+
+The config file will be placed in the following subtree:
+```
+$OS_SPECIFIC_PATH
+├─── test-project-nodejs
+|   └─── Config
```
sindresorhus: This is only correct on Windows. On other systems it's not tested inside the project directory.

test.js
```diff
@@ -126,6 +126,30 @@ test('`configName` option', t => {
...
+	t.true(conf.path.includes('-node'));
+});
+
+test('with `suffix` option set to empty string', t => {
+	const projectSuffix = '';
+	const conf = new Conf({projectSuffix});
+	// Config file's directory root
+	const configRoot = path.resolve(conf.path, '../..');
+	// Extract the config file's parent folder
+	const rootName = path.basename(configRoot);
```
sindresorhus: You should also assert that the `rootName` is correct, as that will only pass on Windows currently.

test.js
```diff
@@ -126,6 +126,30 @@ test('`configName` option', t => {
...
+	// Expect to not have the path
+	t.false(rootName.includes(`-${projectSuffix}`));
+});
+
+test('with `projectSuffix` option set to non-empty string', t => {
+	const projectSuffix = 'new-projectSuffix';
+	const conf = new Conf({projectSuffix});
+	// New option starting from 2.0.0 (excluded)
+	t.true(conf.path.includes(`-${projectSuffix}`));
+});
```
sindresorhus: Thanks for including comprehensive tests. I think the comments can be removed though. They don't really add anything.

readme.md
niktekusho: Thanks for pointing this out. Honestly, I didn't test it on other OSes other than Windows and this does show! 😄 Will fix later.

test.js
niktekusho: Shouldn't test executed on Travis be meant to reflect actual behavior on Linux and Mac as well? I will add it nonetheless if it makes tests more future-proof.

sindresorhus: I think you're misunderstanding what I'm saying. The test is currently faulthy as `t.true(conf.path.includes(`-${projectSuffix}`));` will always pass, since it tests the a negative case but never asserts that what it tests is correct.

You will also have to adjust `const configRoot = path.resolve(conf.path, '../..');` for macOS and Linux.

niktekusho:
I made major updates on the README and on `test.js` (as you required). I now think that maybe the examples on the README are not worth keeping since I don't think they explain the option that much better. Waiting for your review just in case. 😄

PR #54, opened 2018-11-18
Check that require.cache exists

diamondap:
Make sure require.cache exists before trying to alter it. This fixes an 'undefined' error when building with nexe.


sindresorhus:
`require.cache` is guaranteed to exist in Node.js, so you should rather open an issue on `nexe` about this.

https://nodejs.org/api/modules.html#modules_require_cache

diamondap:
You're right. The issue here is how nexe bundles the scripts that it packs into the executable.

PR #55, opened 2018-12-27
Add `onDidAnyChange` method

### LINKED ISSUE #45: On change for the whole config
Hello! Would it be possible to have a `.onChange` handler for the whole config object?

dalgleish:
Fix #45: Hi this is my first contribution so I just attempted to follow the existing convention.
Please let me know if there are any corner cases I didn't consider. I ran tests with node `v6.16.0`, `v8.14.0`, `v10.15.0`.

I moved `isEqual` to a static method and reused it in both `onDidChange` and `onDidAnyChange`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -95,6 +95,39 @@ class Conf {
 		this.store = plainObject();
 	}
 
+	static isEqual(newValue, oldValue) {
+		try {
+			// TODO: Use `util.isDeepStrictEqual` when targeting Node.js 10
+			assert.deepEqual(newValue, oldValue);
+		} catch (_) {
+			return false;
+		}
+
+		return true;
+	}
+
+	onDidAnyChange(callback) {
+		if (typeof callback !== 'function') {
+			throw new TypeError(`Expected \`callback\` to be of type \`function\`, got ${typeof callback}`);
+		}
+
+		let currentValue = this.store;
+
+		const onChange = () => {
+			const oldValue = currentValue;
+			const newValue = this.store;
+
+			const didChange = Conf.isEqual(oldValue, newValue) === false;
+			if (didChange) {
+				currentValue = newValue;
+				callback.call(this, newValue, oldValue);
+			}
+		};
+
+		this.events.on('change', onChange);
+		return () => this.events.removeListener('change', onChange);
+	}
+
 	onDidChange(key, callback) {
 		if (typeof key !== 'string') {
 			throw new TypeError(`Expected \`key\` to be of type \`string\`, got ${typeof key}`);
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -95,6 +95,39 @@ class Conf {
...
+		const onChange = () => {
+			const oldValue = currentValue;
+			const newValue = this.store;
+
+			const didChange = Conf.isEqual(oldValue, newValue) === false;
+			if (didChange) {
+				currentValue = newValue;
+				callback.call(this, newValue, oldValue);
+			}
+		};
```
dalgleish: `onChange` repeats some logic as in `onDidChange`, maybe it could be generalized.

readme.md
```diff
@@ -145,6 +145,12 @@ Delete all items.
 
 Watches the given `key`, calling `callback` on any changes. When a key is first set `oldValue` will be `undefined`, and when a key is deleted `newValue` will be `undefined`.
 
+#### .onDidAnyChange(callback)
+
+`callback`: `(newValue, oldValue) => {}`
+
+Calls `callback` on changes anywhere in the config file. `oldValue` and `newValue` will be the full config object before after the change, respectively. You must compare `oldValue` to `newValue` to find out what changed.
```
brandon93s: before *and* after

index.js
```diff
@@ -95,6 +95,39 @@ class Conf {
 		this.store = plainObject();
 	}
 
+	static isEqual(newValue, oldValue) {
```
brandon93s: This makes it unnecessarily available on the public interface to consumers. This could just be a private helper method in the file, similar to `plainObject` to prevent unintentional usage.

index.js
brandon93s: [dot-prop](https://github.com/sindresorhus/dot-prop) supports [returning the entire object](https://github.com/sindresorhus/dot-prop/blob/42cc3bf8ec8a404bca121a70ea831a9d687582a2/index.js#L24-L26) when a path is excluded:

```js
	get(obj, path, value) {
		if (!isObj(obj) || typeof path !== 'string') {
			return value === undefined ? obj : value;
		}
```

This should pave the path to greatly simplify the two implementations since `onDidAnyChange` will just be `onDidChange` with an empty `key`. To support separate validation, the internals could be extracted into a shared method.

index.js
brandon93s: If refactored into a shared method, extraction of the equal logic may be unnecessary.

sindresorhus:
@dalgleish Bump :)

sindresorhus:
@dalgleish Bump2

Issue #24, opened 2017-09-04
Prevent unsupported JSON values


sindresorhus:
For example, `undefined` is not supported in JSON. We should throw rather than silently coercing it.

acheronfail:
Unsupported values include:

- the value `undefined`
    - `JSON.stringify({ foo: undefined }) ` = `"{}"`
- any `Function`
    - `JSON.stringify({ foo: function () {} })` = `"{}"`
- any `Symbol` (as a key or value)
    - `JSON.stringify({ foo: Symbol('foo') })` = `"{}"`
    - `JSON.stringify({ [Symbol('foo')]: 'foo' })` = `"{}"`


Also, non-enumerable properties aren't supported (from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify)).
```js
const value = Object.create(null, {
  x: {
    value: 'x',
    enumerable: false
  },
  y: {
    value: 'y',
    enumerable: true
  }
});

JSON.stringify(value); // '{"y":"y"}'
```

acheronfail:
I think checking for the first three cases makes sense (check for `undefined`, `typeof x === 'function'` and `typeof x === 'symbol'`. That shouldn't be a difficult change.

Not sure how feasible it really is to check if every value is `enumerable` though, what do you think?

sindresorhus:
> I think checking for the first three cases makes sense (check for undefined, typeof x === 'function' and typeof x === 'symbol'. That shouldn't be a difficult change.

👍 

> Not sure how feasible it really is to check if every value is enumerable though, what do you think?

Yes, doesn't make sense to check for this.

PR #57, opened 2019-01-22
Prevent unsupported JSON values #24

nwmandel:
In reference to this issue here [https://github.com/sindresorhus/conf/issues/24](url). Preventing undefined JSON value to get passed along. Added test case to test file as well.


sindresorhus:
I appreciate the PR, but there's just too much to comment on, and I don't have enough time right now to do that kind of review, so I'm unfortunately going to pass on it and do the change myself.

PR #58, opened 2019-01-31
Add support for JSON schema

yaodingyd:
Fix https://github.com/sindresorhus/electron-store/issues/16

Use epoberezkin/ajv for JSON schema validation

### DIFF
--- a/index.js
+++ b/index.js
@@ -64,8 +64,22 @@ module.exports = class Conf {
 		this._options = options;
 
 		if (options.schema) {
-			const ajv = new Ajv();
-			this._validator = ajv.compile(options.schema);
+			if (typeof options.schema !== 'object') {
+				throw new TypeError('Schema option must be an object.');
+			}
+
+			const ajv = new Ajv({
+				allErrors: true,
+				format: 'full',
+				useDefaults: true,
+				errorDataPath: 'property'
+			});
+			const schema = {
+				type: 'object',
+				properties: options.schema
+			};
+			console.log(schema);
+			this._validator = ajv.compile(schema);
 		}
 
 		this.events = new EventEmitter();
@@ -187,8 +201,9 @@ module.exports = class Conf {
 				} catch (_) {}
 			}
 
+			data = this.deserialize(data);
 			this._validate(data);
-			return Object.assign(plainObject(), this.deserialize(data));
+			return Object.assign(plainObject(), data);
 		} catch (error) {
 			if (error.code === 'ENOENT') {
 				// TODO: Use `fs.mkdirSync` `recursive` option when targeting Node.js 12
--- a/index.js
+++ b/index.js
@@ -63,9 +63,9 @@ module.exports = class Conf {
 
 		this._options = options;
 
-		if (options.schema && options.fileExtension === 'json') {
+		if (options.schema) {
 			const ajv = new Ajv();
-			this.validator = ajv.compile(options.schema);
+			this._validator = ajv.compile(options.schema);
 		}
 
 		this.events = new EventEmitter();
@@ -85,14 +85,16 @@ module.exports = class Conf {
 		}
 	}
 
-	validate(data) {
-		if (this.validator) {
-			const valid = this.validator(data);
-			if (!valid) {
-				const errors = this.validator.errors.reduce((error, {dataPath, message}) =>
-					error + `${dataPath} ${message};`, '');
-				throw new Error('Config is not valid according to schema:' + errors);
-			}
+	_validate(data) {
+		if (!this._validator) {
+			return true;
+		}
+
+		const valid = this._validator(data);
+		if (!valid) {
+			const errors = this._validator.errors.reduce((error, {dataPath, message}) =>
+				error + `${dataPath} ${message};`, '');
+			throw new Error('Config is not valid according to schema: ' + errors);
 		}
 	}
 
@@ -185,7 +187,7 @@ module.exports = class Conf {
 				} catch (_) {}
 			}
 
-			this.validate(JSON.parse(data));
+			this._validate(data);
 			return Object.assign(plainObject(), this.deserialize(data));
 		} catch (error) {
 			if (error.code === 'ENOENT') {
@@ -206,7 +208,7 @@ module.exports = class Conf {
 		// Ensure the directory exists as it could have been deleted in the meantime
 		makeDir.sync(path.dirname(this.path));
 
-		this.validate(value);
+		this._validate(value);
 		let data = this.serialize(value);
 
 		if (this.encryptionKey) {
--- a/index.js
+++ b/index.js
@@ -91,6 +91,7 @@ module.exports = class Conf {
 
 		const fileStore = this.store;
 		const store = Object.assign(plainObject(), options.defaults, fileStore);
+		this._validate(store);
 		try {
 			assert.deepEqual(fileStore, store);
 		} catch (_) {
@@ -100,14 +101,14 @@ module.exports = class Conf {
 
 	_validate(data) {
 		if (!this._validator) {
-			return true;
+			return;
 		}
 
 		const valid = this._validator(data);
 		if (!valid) {
 			const errors = this._validator.errors.reduce((error, {dataPath, message}) =>
-				error + `${dataPath} ${message};`, '');
-			throw new Error('Config is not valid according to schema: ' + errors);
+				error + ` \`${dataPath.slice(1)}\` ${message};`, '');
+			throw new Error('Config schema violation:' + errors.slice(0, -1));
 		}
 	}
 
--- a/index.js
+++ b/index.js
@@ -65,7 +65,7 @@ module.exports = class Conf {
 
 		if (options.schema) {
 			if (typeof options.schema !== 'object') {
-				throw new TypeError('Schema option must be an object.');
+				throw new TypeError('The `schema` option must be an object.');
 			}
 
 			const ajv = new Ajv({
--- a/index.js
+++ b/index.js
@@ -78,7 +78,6 @@ module.exports = class Conf {
 				type: 'object',
 				properties: options.schema
 			};
-			console.log(schema);
 			this._validator = ajv.compile(schema);
 		}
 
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -139,6 +156,8 @@ class Conf {
 				} catch (_) {}
 			}
 
+			this.validate(JSON.parse(data));
```
sindresorhus: No point in doing `JSON.parse` twice.

index.js
```diff
@@ -41,6 +42,11 @@ class Conf {
 			options.cwd = envPaths(options.projectName, {suffix: options.projectSuffix}).config;
 		}
 
+		if (options.schema && options.fileExtension === 'json') {
```
sindresorhus: It should work with any extension.

test.js
```diff
@@ -355,3 +355,36 @@ test('doesn\'t write to disk upon instanciation if and only if the store didn\'t
 	exists = fs.existsSync(conf.path);
 	t.is(exists, true);
 });
+
+test('schema - invalid set', t => {
+	const schema = {
+		$schema: 'http://json-schema.org/draft-07/schema#',
+		type: 'object',
+		properties: {
```
sindresorhus: We should set these top-level meta things for the user. The user should only specify their schema as:

```js
{
	foo: {
		type: 'string'
	}
}
```

readme.md
```diff
@@ -120,6 +120,13 @@ You can pass an empty string to remove the suffix.
 
 For example, on macOS, the config file will be stored in the `~/Library/Preferences/foo-nodejs` directory, where `foo` is the `projectName`.
 
+#### schema
+
+type: `Object`<br>
+Default: `undefined`
+
+JSON Schema to validate your config.
```
sindresorhus: "JSON schema" should link to docs about it.

sindresorhus:
You need to consider how the `defaults` option will work with schema defaults. There's overlap. https://github.com/epoberezkin/ajv#assigning-defaults

sindresorhus:
This PR is a bit thin considering the expected work for the bounty.

sindresorhus:
The readme documentation should have a full realistic example of how it should be used.

sindresorhus:
Also go through all the options in https://github.com/epoberezkin/ajv#validation-and-reporting-options and carefully consider which one to use or not.

index.js
```diff
@@ -56,6 +62,17 @@ class Conf {
 		}
 	}
 
+	validate(data) {
+		if (this.validator) {
```
sindresorhus: Would be more readable with an early-return here.

sindresorhus:
We should at least have these options: https://github.com/wulkano/kap/blob/01a557f6616108423f9e46718d48596a22698480/main/plugin.js#L48-L53

index.js
```diff
@@ -41,6 +42,11 @@ class Conf {
 			options.cwd = envPaths(options.projectName, {suffix: options.projectSuffix}).config;
 		}
 
+		if (options.schema && options.fileExtension === 'json') {
+			const ajv = new Ajv();
+			this.validator = ajv.compile(options.schema);
```
```suggestion
			this._validator = ajv.compile(options.schema);
```

index.js
```diff
@@ -56,6 +62,17 @@ class Conf {
 		}
 	}
 
+	validate(data) {
```
```suggestion
	_validate(data) {
```

test.js
```diff
@@ -355,3 +355,36 @@ test('doesn\'t write to disk upon instanciation if and only if the store didn\'t
...
+			foo: {
+				type: 'string'
+			}
+		}
+	};
+	const conf = new Conf({cwd: tempy.directory(), schema});
+	conf.set('foo', fixture);
+	t.is(conf.get('foo'), fixture);
+	t.throws(() => conf.set('foo', 1), 'Config is not valid according to schema:.foo should be string;');
```
sindresorhus: The error message is missing a space.

sindresorhus:
Can you fix the merge conflict?

yaodingyd:
I go through all the options for ajv:
1. We would only use basic and standard type and formats, no `$data` references, no custom `formats`;
2. We don't use referenced schema options(`schemaId`, `missingRefs`, `extendRefs` or `loadSchema`);
3. Except to `useDefault`, we don't want to modify validated data(`removeAdditional` or `coerceTypes`);
4. We don't use any advanced options except `errorDataPath`, which is good for reporting errors.
So I think to use the same options makes sense.

readme.md
```diff
@@ -225,6 +232,33 @@ const config = new Conf({
 });
 ```
 
+### How can I use `schema` option to validate my config?
+
+Under the hood, JSON Schema validator [ajv](https://github.com/epoberezkin/ajv) is used to validate your config. We use
```
sindresorhus: Don't hard-wrap

readme.md
```diff
@@ -225,6 +232,33 @@ const config = new Conf({
...
+[JSON Schema draft-07](http://json-schema.org/latest/json-schema-validation.html) and support all [validation keywords](https://github.com/epoberezkin/ajv/blob/master/KEYWORDS.md) and [formats](https://github.com/epoberezkin/ajv#formats).
+
+You should define your schema as an object where each key is the name of your data's property and each value is a JSON schema used to validate that property. See more [here](https://json-schema.org/understanding-json-schema/reference/object.html#properties).
+
+Example:
+
+```js
+const schema = {
+	'foo': {
```
sindresorhus: No need to quote the keys.

readme.md
```diff
@@ -225,6 +232,33 @@ const config = new Conf({
...
+		type: 'number',
+		maximum: '100',
+		minimum: '1',
```
sindresorhus: Shouldn't these be numbers?

readme.md
```diff
@@ -225,6 +232,33 @@ const config = new Conf({
...
+	'bar': {
+		type: 'string',
+		format: 'url'
+	}
+};
+
+const config = new Conf({schema});
+```
+
+**Please be aware that your schema `default` will overwrite the corresponding property specified in `defaults` option in Conf.**
```
sindresorhus: It should be mentioned in the `defaults` options that it's moot if you set the `schema` option.

test.js
```diff
@@ -434,3 +434,33 @@ test('`clearInvalidConfig` option - valid data', t => {
...
+test('schema - invalid set', t => {
+	const schema = {
+		foo: {
+			type: 'string'
+		}
+	};
+	const conf = new Conf({cwd: tempy.directory(), schema});
+	conf.set('foo', fixture);
+	t.is(conf.get('foo'), fixture);
+	t.throws(() => conf.set('foo', 1), 'Config is not valid according to schema: .foo should be string;');
```
sindresorhus: The error should be:

```
Config schema violation: `foo` should be string
```

Not that there should be not `.` in front of `foo`. And it should be wrapped in backticks, and no `;` at the end.

test.js
```diff
@@ -434,3 +434,33 @@ test('`clearInvalidConfig` option - valid data', t => {
 	conf.set('foo', 'bar');
 	t.deepEqual(conf.store, {foo: 'bar'});
 });
+
+test('schema - should be an object', t => {
+	const schema = 'object';
+	t.throws(() => new Conf({cwd: tempy.directory(), schema}), 'Schema option must be an object.');
```
sindresorhus: Don't use inline arrow function for `t.throws`.

test.js
```diff
@@ -434,3 +434,33 @@ test('`clearInvalidConfig` option - valid data', t => {
...
+		foo: {
+			type: 'string'
+		}
+	};
+	const conf = new Conf({cwd: tempy.directory(), schema});
+	conf.set('foo', fixture);
+	t.is(conf.get('foo'), fixture);
+	t.throws(() => conf.set('foo', 1), 'Config is not valid according to schema: .foo should be string;');
+});
+
```
sindresorhus: Can you add a test for the error when there are multiple violations, to ensure it shows both errors and that it shows it correctly formatted.

readme.md
sindresorhus: Something like this:

> If you use the `schema` option, you can just set the `default` key there for values instead of using this option.

index.js
```diff
@@ -79,6 +99,19 @@ module.exports = class Conf {
 		}
 	}
 
+	_validate(data) {
+		if (!this._validator) {
+			return true;
```
sindresorhus: Why are you returning `true` here? You're not using it anywhere.

readme.md
```diff
@@ -225,6 +232,33 @@ const config = new Conf({
 });
 ```
 
+### How can I use `schema` option to validate my config?
```
sindresorhus: This should just be documented where the `schema` option is defined.

test.js
```diff
@@ -434,3 +434,33 @@ test('`clearInvalidConfig` option - valid data', t => {
...
+		foo: {
+			type: 'string'
+		}
+	};
+	const cwd = tempy.directory();
+
+	const conf = new Conf({cwd, schema});
+	fs.writeFileSync(path.join(cwd, 'config.json'), JSON.stringify({foo: 1}));
+	t.throws(() => conf.get('foo'), 'Config is not valid according to schema: .foo should be string;');
+});
```
sindresorhus: The `schema` option needs more tests:

- Valid config with the schema option.
- Using `default` in the schema.
- `default` in schema overwriting property in `defaults` option.
- Advanced schema with many different types and combinations of attributes.
- What else?...

yaodingyd:
I was wrong about the overwrite order before: actually we apply Conf's defaults first, thus schema's default would not be used because value has already be set. I have updated doc accordingly.

readme.md
```diff
@@ -145,6 +147,38 @@ You can pass an empty string to remove the suffix.
...
+		minimum: 1,
+		default: 50
+	},
+	bar: {
+		type: 'string',
+		format: 'url'
+	}
+};
+
+const config = new Conf({schema});
```
sindresorhus: Can you include an example of setting an invalid value and the error message, so people get a feel of how it works?

readme.md
```diff
@@ -225,7 +259,6 @@ const config = new Conf({
 });
 ```
 
-
```
sindresorhus: Don't do unrelated changes.

sindresorhus:
Can you add the option to https://github.com/sindresorhus/conf/blob/master/index.d.ts and https://github.com/sindresorhus/conf/blob/master/index.test-d.ts ?

And fix the merge conflict.

sindresorhus:
Looks good now :)

PR #63, opened 2019-03-31
feat(version)

robertgrzonka:
Adapted conf because of deprecations:
* add `util.isDeepStrictEqual()` for Node equal or above version 10 (`assert.deepEqual()` is deprecated)
* add `fs.mkdirSync()` for Node equal or above version 10

### DIFF
--- a/index.js
+++ b/index.js
@@ -139,10 +139,14 @@ class Conf {
 		const onChange = () => {
 			const oldValue = currentValue;
 			const newValue = this.get(key);
+			const nodeVersion = process.version.slice(1)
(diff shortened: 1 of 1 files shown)


robertgrzonka:
@sindresorhus if I understood something wrong or worse—did something wrong I am ready for your feedback :)

sindresorhus:
There's really no benefit to doing this yet. The whole point of the TODO comments it to remind me *when* we do target those Node.js versions. For now, using the `make-dir` module and the assert API works just as well.

PR #64, opened 2019-03-31
Breaking: Update dependencies, refactor TypeScript definition to CommonJS compatible export

BendingBender:
**Breaking change** for the same reasons as `PCancelable`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -207,4 +207,3 @@ class Conf {
 }
 
 module.exports = Conf;
-module.exports.default = Conf;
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -207,4 +207,3 @@ class Conf {
 }
 
 module.exports = Conf;
-module.exports.default = Conf;
```
sindresorhus: Is this a breaking change? I'm asking since this is being removed, but the PR title doesn't make it clear.

BendingBender: Sorry, I've written this in the description, not in the title. Will change.

sindresorhus: Oops, I missed that.

sindresorhus:
I will not do a new release here yet. I have some other changes planned before the next major bump.

PR #66, opened 2019-04-11
Add `.onDidAnyChange`

### LINKED ISSUE #45: On change for the whole config

yaodingyd:
fix #45.


sindresorhus:
Looks good :) Can you add this to `electron-store` too?

yaodingyd:
Sure! Just to be clear that would be update `Conf` version (after this change is published) and add option to readme file, right?

PR #67, opened 2019-04-11
Add `accessPropertiesByDotNotation` option

### LINKED ISSUE #44: Add feature to disable dot-notation
[Issuehunt badges](https://issuehunt.io/r/sindresorhus/conf/issues/44)

Hi! I'm working on a project that uses reverse-domain naming for namespaces, and we need to store some data with the namespace as the key. Where we'd expect 
```JSON
{
    "cc.xylobol.test": {
        "foo": "bar"
    }
}
```
we end up with 
```JSON
{
    "cc": {
        "xylobol": {
            "test": {
                "foo": "bar"
            }
        }
    }
}
```
instead due to the dot-notation. Is there a pre-existing way to disable this? If not, could it be added as a feature?

Thank you in advanced.
Xylobol

> There is a $40.00 open bounty on this issue. Add more on [Issuehunt](https://issuehunt.io/r/sindresorhus/conf/issues/44).
> - Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/conf/) to discover more funded issues.
> - Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on Issuehunt to raise funds.

yaodingyd:
fix #44

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -127,7 +127,7 @@ declare namespace Conf {
 		readonly projectSuffix?: string;
 
 		/**
-		Accessing nested properties by dot nonation. For example:
+		Accessing nested properties by dot notation. For example:
 
 		@example
 		```
--- a/index.d.ts
+++ b/index.d.ts
@@ -161,7 +161,6 @@ declare namespace Conf {
 
 		@default true
 		*/
-
 		readonly accessPropertiesByDotNotation?: boolean;
 	}
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -159,7 +159,7 @@ declare namespace Conf {
 		//=> '🦄'
 		```
 
-		@default 'true'
+		@default true
 		*/
 
 		readonly accessPropertiesByDotNotation?: boolean;
--- a/index.d.ts
+++ b/index.d.ts
@@ -125,6 +125,15 @@ declare namespace Conf {
 		@default 'nodejs'
 		*/
 		readonly projectSuffix?: string;
+
+		/**
+		Disable dot-notation for accessing nested properties, treat the whole string as one key.
+
+		@default 'false'
+		*/
+
+		readonly disableDotNotation?: boolean;
+
 	}
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -127,12 +127,42 @@ declare namespace Conf {
 		readonly projectSuffix?: string;
 
 		/**
-		Disable dot-notation for accessing nested properties, treat the whole string as one key.
+		Accessing nested properties by dot nonation. For example:
 
-		@default 'false'
+		@example
+		```
+		const config = new Conf();
+
+		config.set({
+			foo: {
+				bar: {
+					foobar: '🦄'
+				}
+			}
+		});
+
+		console.log(config.get('foo.bar.foobar'));
+		//=> '🦄'
+		```
+
+		Alternatively, you can set this option to `false` so the whole string would be treated as one key.
+
+		@example
+		```
+		const config = new Conf({accessPropertiesByDotNotation: false});
+
+		config.set({
+			`foo.bar.foobar`: '🦄'
+		});
+
+		console.log(config.get('foo.bar.foobar'));
+		//=> '🦄'
+		```
+
+		@default 'true'
 		*/
 
-		readonly disableDotNotation?: boolean;
+		readonly accessPropertiesByDotNotation?: boolean;
 
 	}
 }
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -188,6 +188,13 @@ You can pass an empty string to remove the suffix.
 
 For example, on macOS, the config file will be stored in the `~/Library/Preferences/foo-nodejs` directory, where `foo` is the `projectName`.
 
+#### disableDotNotation
```
sindresorhus: Double-negative option names are not optimal. Would be better with `dotNotation`.

readme.md
```diff
@@ -188,6 +188,13 @@ You can pass an empty string to remove the suffix.
...
+
+Type: `boolean`<br>
+Default: `false`
+
+Disable dot-notation for accessing nested properties, treat the whole string as one key.
```
sindresorhus: Add an example. Might not be clear to users what `dot-notation` means.

index.js
```diff
@@ -110,8 +110,14 @@ class Conf {
...
 
+	_getkey(key) {
+		const escapeDot = value => value.replace(/\./g, '\\\\.');
+
+		return this._options.disableDotNotation ? escapeDot(key) : key;
+	}
+
 	get(key, defaultValue) {
-		return dotProp.get(this.store, key, defaultValue);
+		return dotProp.get(this.store, this._getkey(key), defaultValue);
```
sindresorhus: Instead of escaping the dot, simply don't use `dotProp` when dot-notation is turned off.

sindresorhus:
I'm not super happy about the name `dotNotation`. Can you list some other name ideas for the option?

yaodingyd:
Most explicit would be `accessPropertyByDotDonation`

index.d.ts
```diff
@@ -125,6 +125,45 @@ declare namespace Conf {
...
+
+		config.set({
+			`foo.bar.foobar`: '🦄'
+		});
+
+		console.log(config.get('foo.bar.foobar'));
+		//=> '🦄'
+		```
+
+		@default 'true'
```
sindresorhus:
```suggestion
		@default true
```

index.d.ts
```diff
@@ -125,6 +125,45 @@ declare namespace Conf {
 		@default 'nodejs'
 		*/
 		readonly projectSuffix?: string;
+
+		/**
+		Accessing nested properties by dot nonation. For example:
```
sindresorhus: Typo

test.js
```diff
@@ -11,6 +11,7 @@ const fixture = '🦄';
 
 test.beforeEach(t => {
 	t.context.conf = new Conf({cwd: tempy.directory()});
+	t.context.config = new Conf({cwd: tempy.directory(), accessPropertiesByDotNotation: false});
```
sindresorhus: It's weird that this is called `config` and the other one `conf`. Be more specific.

index.d.ts
```diff
@@ -125,6 +125,45 @@ declare namespace Conf {
...
+
+		console.log(config.get('foo.bar.foobar'));
+		//=> '🦄'
+		```
+
+		@default true
+		*/
+
+		readonly accessPropertiesByDotNotation?: boolean;
+
```
sindresorhus:
```suggestion
```

index.d.ts
```diff
@@ -125,6 +125,45 @@ declare namespace Conf {
...
+			`foo.bar.foobar`: '🦄'
+		});
+
+		console.log(config.get('foo.bar.foobar'));
+		//=> '🦄'
+		```
+
+		@default true
+		*/
+
```
sindresorhus:
```suggestion
```

index.d.ts
```diff
@@ -125,6 +125,45 @@ declare namespace Conf {
...
+		config.set({
+			`foo.bar.foobar`: '🦄'
+		});
+
+		console.log(config.get('foo.bar.foobar'));
+		//=> '🦄'
+		```
+
+		@default true
```
sindresorhus: This should be before the example.

index.d.ts
```diff
@@ -125,6 +125,45 @@ declare namespace Conf {
 		@default 'nodejs'
 		*/
 		readonly projectSuffix?: string;
+
+		/**
+		Accessing nested properties by dot notation. For example:
```
sindresorhus:
```suggestion
		Access nested properties by dot notation.
```

sindresorhus:
I feel like the option is a bit too verbose, but I can't really think of anything else, and at least it's clearer than `dotNotation`.

sindresorhus:
A new release is out. Can you document this PR and the other one in `electron-store` too? https://github.com/sindresorhus/conf/releases/tag/v4.1.0

yaodingyd:
sure, on it

PR #68, opened 2019-04-12
Minor code refactor

Hongarc:
This way will make `get` and `set` use the same algorithm


index.js
```diff
@@ -13,6 +13,7 @@ const writeFileAtomic = require('write-file-atomic');
 const Ajv = require('ajv');
 
 const plainObject = () => Object.create(null);
+const algorithm = 'aes-256-cbc';
```
```suggestion
const encryptionAlgorithm = 'aes-256-cbc';
```

Hongarc: I changed it, thank you for suggesting.

PR #72, opened 2019-05-23
Replace deprecated crypto methods

MattMattV:
I wanted to fix https://github.com/sindresorhus/electron-store/issues/67, but the fix had to be done here.

Since the README states :

> Note that this is not intended for security purposes, since the encryption key would be easily found inside a plain-text Node.js app.

I permitted myself to use a non-random IV.

<details>
<summary>Test results</summary>
Node 11.15.0 

Test results on Node 11.15.0

Node 10.15.3 

Test results on Node 10.15.3

Node 8.16.0

Test results on Node 8.16.0
</details>

Also I would be very grateful if you could indicate me the next step so I can help you resolve https://github.com/sindresorhus/electron-store/issues/67

Thanks in advance !

### DIFF
--- a/index.js
+++ b/index.js
@@ -211,7 +211,10 @@ class Conf {
 
 			if (this.encryptionKey) {
 				try {
-					const decipher = crypto.createDecipher(encryptionAlgorithm, this.encryptionKey);
+					const iv = Buffer.alloc(16);
+					const key = crypto.pbkdf2Sync(this.encryptionKey, iv.toString(), 10000, 32, 'sha512');
+
+					const decipher = crypto.createDecipheriv(encryptionAlgorithm, key, iv);
 					data = Buffer.concat([decipher.update(data), decipher.final()]);
 				} catch (_) {}
 			}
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -211,7 +211,10 @@ class Conf {
 
 			if (this.encryptionKey) {
 				try {
-					const decipher = crypto.createDecipher(encryptionAlgorithm, this.encryptionKey);
+					const iv = Buffer.alloc(16);
+					const key = crypto.pbkdf2Sync(this.encryptionKey, iv.toString(), 10000, 32, 'sha512');
```
sindresorhus: I think this will break existing uses where the `encryptionKey` is not using `pbkdf2Sync`.

MattMattV: According to my tries and by https://github.com/nodejs/node/issues/16746 it may impossible to make this upgrade without any harm since `key` patrameter in `createCipherkv` method require fixed length, so any previous key that does not match `createCipherIv` criteria will produce an error :confused:

sindresorhus:
Closing this in favor of https://github.com/sindresorhus/conf/pull/76

PR #73, opened 2019-06-05
Improve TypeScript typings

TheAifam5:
This PR extends the typings definition for better usability and error checking.
It asserts the class property names and its type for all functions.


index.d.ts
```diff
@@ -1,5 +1,5 @@
 /// <reference types="node"/>
-import {JSONSchema} from 'json-schema-typed';
+import { JSONSchema } from "json-schema-typed";
```
sindresorhus: Don't do unrelated changes.

index.test-d.ts
```diff
@@ -1,57 +1,77 @@
...
+const conf = new Conf<UnicornFoo>();
+new Conf<UnicornFoo>({
 	defaults: {
-		foo: 'bar',
-		unicorn: 'rainbow'
+		foo: "bar",
+		unicorn: false
+	}
+});
+new Conf<UnicornFoo>({ configName: "" });
```
sindresorhus: Use single-quotes

tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+	"compilerOptions": {
+		"module": "commonjs",
+		"strict": true,
+		"downlevelIteration": true
+	},
+	"exclude": [
+		"node_modules"
+	]
+}
```
sindresorhus: This file is not needed.

sindresorhus:
Thanks for working on improving the types. Can you describe what you changed and why? How is it better?

sindresorhus:
@TheAifam5 Friendly bump :)

sindresorhus:
Despite lack of response, the typings here are much better than master, so I'll merge this and clean it up. Thanks, @TheAifam5 :)

jsynowiec:
@sindresorhus `Partial<t>` is not the same as `{[key: string]: T}`. I don't know if this was intended, but it makes all the properties optional, hence could be a breaking change to some projects.


```
/**
 * Make all properties in T optional
 */
type Partial<T> = {
    [P in keyof T]?: T[P];
};
```

sindresorhus:
@jsynowiec Can you be more specific about which `Partial` call is problematic?

TheAifam5:
@sindresorhus 

The `Partial` type makes all properties optional.
What you have (in the probject):
```
interface MyConfig {
  alwaysPresent: string;
  willBeSetMaybeLater?: string;
}
```

What the will have:
```
interface MyConfig {
  alwaysPresent?: string;
  willBeSetMaybeLater?: string;
}
```

Let's say, you sure that you will have the `alwaysPresent`, but the library tells you that IS optional and can be empty - which is wrong, you will need to check the presence first and then you can access the value.

Just remove that `Partial` thing from the code (I think you can remove it everywhere).

jsynowiec:
Well, all of them :-) Maybe change `Partial<T>` to `{[P in keyof T]: T[P]}` as it's more explicit but doesn't change the properties on `T`?

TheAifam5:
@jsynowiec Nope. But you can do just `T` and you will get the same results. If you want to be more explicit, then you can do something like:
```class/interface XYZ<T extends object>```.

jsynowiec:
I know that, but we started this discussion from my question about introducing this change in the first place ;-) `Partial` **is** what I wrote in the previous post + the optional parameters.

Right now, this PR merged and released in `6.0.0` broke few of my projects and I really looked for the migrations feature 😞

sindresorhus:
I missed that the `store` is `Partial` too.

@TheAifam5 Why did you make the `store` `Partial`?

---

https://github.com/sindresorhus/conf/blob/bb24cfe32d39617a4b3983a92bd81596bd87c312/index.d.ts#L258 has to be `Partial`, but I think that's the only one.

TheAifam5:
I made a PR to fix that issues.

PR #76, opened 2019-06-16
Replace deprecated `crypto.createDecipher()` and `crypto.createCipher()` Node.js methods

popod:
Fix sindresorhus/electron-store#67

It is better than #72 in this way:
- It doesn't break existing encrypted data
- The data encryption is secure and could be used for security purposes


sindresorhus:
> It doesn't break existing encrypted data

Can you explain how it manages that?

> The data encryption is secure and could be used for security purposes

That's meaningless as the decryption key can be easily found in the JS code of the app.

popod:
> 1. It doesn't break existing encrypted data

The new encrypted data contain the IV vector in it. Data are stored like this: `<the IV vector>:<the config data>`. By this way, we can know (on the decrypt process) if the data had beeen encrypted with `crypto.createCipher()` (no IV vector is stored with the data) or `crypto.createCipher()`. In the other PR, already existed data (encrypted with `crypto.createCipher()` without the iv vector) cannot be decrypted. See the code to understand.

> 2. The data encryption is secure and could be used for security purposes

Yes, but this is no more because of sindresorhus/conf. Now the security is assured by this package. This is now the developer of the application which shoud take care of the password security if it want to use this package for security purposes. If the key used to decrypt the data is stored with `keytar` or is a password entered by the user, this is secure.

Is it more understandable? ;)

popod:
Hi @sindresorhus,

Is my proposal okey for you ? Do you require some changes or additional clarification ?

Thanks

sindresorhus:
Yes, makes sense. I think indeed this is the best solution. Can you add a test that ensures a config encrypted with the existing version will be correctly decrypted with this version? Just commit a small config fixture encrypted with the current version (in master).

popod:
@sindresorhus done.

Do you think we should update this too in `readme.md` ?

> Note that this is not intended for security purposes, since the encryption key would be easily found inside a plain-text Node.js app.

sindresorhus:
Sure

popod:
@sindresorhus is all okey for you now ?

test.js
```diff
@@ -360,6 +360,12 @@ test('encryption - corrupt file', t => {
 	t.is(after.get('foo'), undefined);
 });
 
+test('decription migration to IV', t => {
+	// The encrypted_config.json contain '{'unicorn', '🦄'}' encrypted with conf@4.1.0 and password 'abcd1234'
```
sindresorhus: It cannot contain `{'unicorn', '🦄'}` as it's not valid JSON. Please include *exactly* what it contains.

sindresorhus:
Filename `test/encrypted_config.json` => `test/config-encrypted-with-conf-4-1-0.json`

index.js
```diff
@@ -229,8 +229,16 @@ class Conf {
 
 			if (this.encryptionKey) {
 				try {
-					const decipher = crypto.createDecipher(encryptionAlgorithm, this.encryptionKey);
-					data = Buffer.concat([decipher.update(data), decipher.final()]);
+					// Check if an IV has been used to encrypt the data
+					if (data.slice(16, 17).toString() === ':') {
+						const iv = data.slice(0, 16);
+						const pass = crypto.pbkdf2Sync(this.encryptionKey, iv.toString(), 10000, 32, 'sha512');
```
sindresorhus: Don't use abbreviations. `pass` => `password`, `iv` => `initializationVector`

test.js
```diff
@@ -360,6 +360,12 @@ test('encryption - corrupt file', t => {
 	t.is(after.get('foo'), undefined);
 });
 
+test('decription migration to IV', t => {
```
sindresorhus: typo

readme.md
```diff
@@ -129,9 +129,9 @@ The only use-case I can think of is having the config located in the app directo
 Type: `string` `Buffer` `TypedArray` `DataView`<br>
 Default: `undefined`
 
-Note that this is **not intended for security purposes**, since the encryption key would be easily found inside a plain-text Node.js app.
+This can be used to secure sensitive data **if** the encryption key is stored in a secure manner (not plain-text) in the Node.js app.
```
sindresorhus: I think you should write a little bit more on how to store the encryption key in a secure manner. 

You mentioned some of this in the PR comments:

> If the key used to decrypt the data is stored with keytar or is a password entered by the user (stored in a variable by example), this is secure.

test.js
popod: Sorry this is a typo.. I will say `{'unicorn': '🦄'}` which is created with `config.set('unicorn', '🦄')`.

index.js
popod: corrected

sindresorhus:
A new release is out. Can you add this to electron-store too in form of a dependency bump and readme updates?

https://github.com/sindresorhus/conf/releases/tag/v5.0.0

popod:
Yes, I will do it. Thanks for reviews and merge ;)

PR #78, opened 2019-06-19
chore(Crypto): use createCipheriv and createDecipheriv

KillerCodeMonkey:
Use non deprecated crypto methods.
Extend typings and readme

fixes https://github.com/sindresorhus/electron-store/issues/67

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,6 @@
 /// <reference types="node"/>
 import {JSONSchema} from 'json-schema-typed';
+import { KeyObject } from 'crypto';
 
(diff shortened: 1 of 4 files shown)


sindresorhus:
There's already a PR for this => https://github.com/sindresorhus/conf/pull/76

KillerCodeMonkey:
yeah, but i find the implementation a little bit strange, to generate the iv in the conf library, because if a user passes an encryption key the iv should be passed as well.

And there was a little typings fix --> encryptionKey can be a crypto "KeyObject" as well.

PR #82, opened 2019-07-11
fix: do not use chown on write - system default permissions is enough

develar:
Fixes https://github.com/sindresorhus/electron-store/issues/75

### DIFF
--- a/index.js
+++ b/index.js
@@ -274,7 +274,7 @@ class Conf {
 			data = Buffer.concat([initializationVector, Buffer.from(':'), cipher.update(Buffer.from(data)), cipher.final()]);
 		}
 
-		writeFileAtomic.sync(this.path, data);
+		writeFileAtomic.sync(this.path, data, {chown: false, mode: false});
 		this.events.emit('change');
 	}
 


sindresorhus:
Tests are failing https://travis-ci.org/sindresorhus/conf/jobs/557177553

develar:
Deeper and deeper... it is bug in write-file-atomic — https://github.com/npm/write-file-atomic/pull/51

anujdatar:
I actually tried locally implementing changes from this PR (https://github.com/sindresorhus/conf/pull/82) and the one from https://github.com/npm/write-file-atomic/pull/51. Still doesn't work with Snap, Same EPERM error

sindresorhus:
@develar Would you be able to publish your fork of `write-file-atomic` to npm so we can depend on that one in the meantime?

sindresorhus:
@anujdatar There might be other things in your dependency tree using a different version of `write-file-atomic`.

anujdatar:
@sindresorhus I actually tried a few fixes mentioned here and elsewhere, couldn't get `write-file-atomic` to work with strict snaps. then I just resorted to using `fs.writeFileSync()`.. I know it's not ideal, but I am not writing production apps, so it works for me. I used your `conf` module as reference to write a simpler version that uses `fs.writeFileSync`.. [`appconfig`](https://github.com/anujdatar/appconfig)

sindresorhus:
Could we detect if it’s a Snap app and use fs.writeFileSync then?

anujdatar:
maybe an environment variable or something?

jwheare:
process.env.SNAP works

anujdatar:
@jwheare yeah, thats what I was thinking.

Or, maybe you can just add something to the readme saying Snaps might need to fork and replace `write-file-atomic` with `fs.writeFileSync`. Because, it's not like you need to use Snap packages, you can just use AppImages if you want one app to work across distributions. This was just something I ran into while playing around. I haven't tried Flatpaks yet, don't know if there's anything there.

sindresorhus:
I have added a temporary workaround in the latest version: https://github.com/sindresorhus/conf/releases/tag/v6.0.0

tarbear123:
> I have added a temporary workaround in the latest version: https://github.com/sindresorhus/conf/releases/tag/v6.0.0

do i download the file and put in the onenote folder?

sindresorhus:
Closing for now as it doesn't look like the required issue will be fixed any time soon and we already have a workaround in place.

Issue #74, opened 2019-06-12
Add store migrations


sindresorhus:
[<img alt="Issuehunt badges" src="https://img.shields.io/badge/IssueHunt-%2480%20Funded-%2300A156.svg" />](https://issuehunt.io/r/sindresorhus/conf/issues/74)

See https://github.com/sindresorhus/conf/pull/49 for an initial attempt that went stale. Please take into account all the feedback given there if you intend to take on the issue.

Part of this task is also to bring the docs over to `electron-store` to resolve https://github.com/sindresorhus/electron-store/issues/19.

---

> Note: This is not an easy issue. It requires you to have good JavaScript and Node.js experience. I expect you to do a good job. Instead of asking to many questions, present solutions. The point of a issue bounty is to give me less work, not more. 🙌

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

- [<img src='https://avatars3.githubusercontent.com/u/44827199?v=4' alt='issuehunt' width=24 height=24> issuehunt](https://issuehunt.io/u/issuehunt) ($80.00)
### Submitted pull Requests
- [#83 Adds support for migrations](https://issuehunt.io/r/sindresorhus/conf/pull/83)
---

### Tips

- Checkout the [Issuehunt explorer](https://issuehunt.io/r/sindresorhus/conf/) to discover more funded issues.
- Need some help from other developers? [Add your repositories](https://issuehunt.io/r/new) on IssueHunt to raise funds.
---
IssueHunt has been backed by the following sponsors. [Become a sponsor](https://issuehunt.io/membership/members)
</details>

rafaelramalho19:
I'm more than happy to pick this up.

Wish me luck 🤞

PR #83, opened 2019-07-22
Add support for migrations

rafaelramalho19:
Fixes https://github.com/sindresorhus/conf/issues/74

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#74: Add store migrations](https://issuehunt.io/repos/61670221/issues/74)
---

</details>


index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
...
+
+		const migrationsToRun = olderVersionsWithMigrations.map(version => migrations[version]);
+
+		migrationsToRun.forEach(migration => migration(this));
+
+		this.set(MIGRATION_KEY, versionToMigrate);
+	}
+
+	/**
+	 * Specifies if a version is suitable to run its migrations.
```
rafaelramalho19: @sindresorhus We didn't have any jsdocs in the remaining file, however, I found it appropriate to add internal documentation to this function.


If you prefer, I can either:

1. Delete this jsdoc
2. Add jsdocs to all the other functions.

I'm fine with either option, and I don't mind the work of last option 😄

index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
 		}
 	}
 
+	_migrate(migrations = {}, versionToMigrate = '0.0.0') {
+		const MIGRATION_KEY = '__conf-migrated-version__';
+
+		const previousMigratedVersion = this.get(MIGRATION_KEY, '0.0.0');
+
+		const olderVersionsWithMigrations = Object.keys(migrations)
```
satazor: Nifty pick: perhaps a better name for this variable should be `olderVersions` or `migrationVersions` as it contains an array of versions that need to be migrated?

index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
...
+			.filter(version => this._shouldVersionPerformMigration(version, previousMigratedVersion, versionToMigrate))
+			.sort(semver.compare);
+
+		const migrationsToRun = olderVersionsWithMigrations.map(version => migrations[version]);
+
+		migrationsToRun.forEach(migration => migration(this));
```
satazor: What happens if one of the migration functions fail? Should we have to worry about this by having some sort of "restore snapshot on error"? Or should we simply say that the migration functions can run more than once for the same version, meaning they must be idempotent?

satazor: I'm more inclined to the second.

index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
 		}
 	}
 
+	_migrate(migrations = {}, versionToMigrate = '0.0.0') {
```
satazor: I don't see the value of having default parameters on this function.

index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
...
+	 * @param {string} previousMigratedVersion The version matching the last migration that ocurred.
+	 * @param {string} versionToMigrate The version that specifies what to migrate to (a.k.a the current version).
+	 * @returns {boolean} If the version is suitable or not.
+	 */
+	_shouldVersionPerformMigration(version, previousMigratedVersion, versionToMigrate) {
+		if (semver.lte(version, previousMigratedVersion)) {
+			return false;
+		}
+
+		if (semver.gt(version, versionToMigrate)) {
```
satazor: Shouldn't this be `semver.gte`?

index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
...
+	 * Specifies if a version is suitable to run its migrations.
+	 * To be suitable, a version should be comprehended between the previous migrated version and the version desired to migrate.
+	 *
+	 * @param {string} version The version that's being tested to be suitable or not.
+	 * @param {string} previousMigratedVersion The version matching the last migration that ocurred.
+	 * @param {string} versionToMigrate The version that specifies what to migrate to (a.k.a the current version).
+	 * @returns {boolean} If the version is suitable or not.
+	 */
+	_shouldVersionPerformMigration(version, previousMigratedVersion, versionToMigrate) {
+		if (semver.lte(version, previousMigratedVersion)) {
```
satazor: Nifty pick: `_shouldVersionPerformMigration` -> `_shouldPerformMigration`

index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
...
+	/**
+	 * Specifies if a version is suitable to run its migrations.
+	 * To be suitable, a version should be comprehended between the previous migrated version and the version desired to migrate.
+	 *
+	 * @param {string} version The version that's being tested to be suitable or not.
+	 * @param {string} previousMigratedVersion The version matching the last migration that ocurred.
+	 * @param {string} versionToMigrate The version that specifies what to migrate to (a.k.a the current version).
+	 * @returns {boolean} If the version is suitable or not.
+	 */
+	_shouldVersionPerformMigration(version, previousMigratedVersion, versionToMigrate) {
```
satazor: Nifty pick: _shouldVersionPerformMigration -> _shouldPerformMigration

readme.md
```diff
@@ -1,5 +1,4 @@
 # conf [![Build Status](https://travis-ci.org/sindresorhus/conf.svg?branch=master)](https://travis-ci.org/sindresorhus/conf)
-
```
satazor: Can you keep this line?

index.js
rafaelramalho19: I think I should actually call it newerVersions since they actually represent a migration with a version bigger than the current one 🤔

index.js
rafaelramalho19: Nice catch!

So what you're saying in the first example is:

```
1.0.0 -> success (set migration_key to 1.0.0)
1.0.1 -> success (set migration_key to 1.0.1)
1.0.2 -> fails (throw error, leave migration_key on 1.0.1 and restore previous changes to the store)
```

and in the second


```
1.0.0 -> success 
1.0.1 -> success 
1.0.2 -> fails (throw error, leave migration_key on 0.0.0 and restore previous changes to the store)
```

?

index.js
rafaelramalho19: Makes perfect sense, will change 😄

rafaelramalho19: Nifty question: Should I also update the first description line of the documentation to reflect the new name?

e.g. (Specifies if a migration is suitable to run according to its version)

index.js
rafaelramalho19: Version = version currently being iterated
VersionToMigrate = version the client has

Let's say we have these migrations:

```js
const migrations = {
    '1.0.0': () => initialMigration
    '1.0.1': () => finalMigration
}
```

And the user version is `1.0.1`


Do we want to run :
1. `initialMigration` + `finalMigration` 
or
2.  Just `initialMigration`?

Your suggestions results in option #2, while the current code results in option #1.

readme.md
rafaelramalho19: Yes, of course. I have no idea why this got deleted 😕

index.js
rafaelramalho19: The first one I could agree with you since it's only possible to arrive at this code block with a positive value in the migrations object.

Do you think the last one should be removed as well?  It's possible for the options.projectVersion to be undefined, which would cause this code to break.
One could also argue that if the migrations object is specified, but the projectVersion is not, we should throw an error.

Thoughts on this?

index.js
satazor: You are almost correct, though in the second example:

- The migration_key would be updated to 1.0.1.
- No changes would be restored. The handler for `1.0.2` would be run again in the future, meaning that it should be idempotent. By idempotent, I mean ensuring that the outcome of the overall migration for that version is the same, regardless of a previous partly successful run.

The second approach is much easier to implement. The first example is difficult to implement because of the "restore previous changes". What happens if we fail to restore the previous changes too? Since we are not dealing with a store that supports transactions, this would be almost impossible to implement correctly on failures.

rafaelramalho19: Storing the values before the mutations could be possible by using the `onDidChange` and using the `callback` to temporarily store the store state.

I do agree that your suggestion is simpler and works fine (sorry for miss-understanding your text), so I don't think there's a need to over-complicate with the "restore state". 

I will add the code to allow for this and also update the documentation in both the repositories. Thank you for clearing that up!

index.js
satazor: Please ignore my comment, the conditions seem to be correct. I guess it's late and I need to sleep 😴

index.js
```diff
@@ -112,6 +122,43 @@ class Conf {
...
+	/**
+	 * Specifies if a version is suitable to run its migrations.
+	 * To be suitable, a version should be comprehended between the previous migrated version and the version desired to migrate.
+	 *
+	 * @param {string} version The version that's being tested to be suitable or not.
+	 * @param {string} previousMigratedVersion The version matching the last migration that ocurred.
+	 * @param {string} versionToMigrate The version that specifies what to migrate to (a.k.a the current version).
+	 * @returns {boolean} If the version is suitable or not.
+	 */
+	_shouldPerformMigration(version, previousMigratedVersion, versionToMigrate) {
```
satazor: Just a small suggestion, what if we rename `version` to `canditateVersion` or something similar to make it more obvious?

index.js
satazor: Agree!

index.js
satazor: I think it's fine, although I think @sindresorhus will ask you to remove the JSDoc to keep consistency since we have none atm. But I will leave that to him.

index.js
rafaelramalho19: It's fine, review it when you have some coffee in your system 😄

index.js
```diff
@@ -112,6 +122,52 @@ class Conf {
...
+			.sort(semver.compare);
+
+		newerVersions.forEach(version => {
+			try {
+				const migration = migrations[version];
+				migration(this);
+
+				this.set(MIGRATION_KEY, version);
+			} catch (error) {
+				console.error('Something went wrong during the migration!');
```
rafaelramalho19: What you guys think about providing information with console.error?

If you agree with this, should we add a stub in the tests to prevent this from being printed?

image

index.js
rafaelramalho19: Sure 👍

index.js
satazor: I'm fine with the `console.error` or `console.warn`, but I will leave that to @sindresorhus

index.js
rafaelramalho19: Done, please review again when you have time 😄

index.js
```diff
@@ -112,6 +122,52 @@ class Conf {
...
+		const newerVersions = Object.keys(migrations)
+			.filter(candidateVersion => this._shouldPerformMigration(candidateVersion, previousMigratedVersion, versionToMigrate))
+			.sort(semver.compare);
+
+		newerVersions.forEach(version => {
+			try {
+				const migration = migrations[version];
+				migration(this);
+
+				this.set(MIGRATION_KEY, version);
```
satazor: Small remark: because we are setting the version here, we would have a duplicate set operation on line 147 for the last entry of the `newerVersions` array. Moreover, if there `newerVersions` array is empty, the set operation on line 147 will always run.

To avoid both issues we can:

1. Set `previousMigratedVersion = version`  here and add a condition on line 147 that only performs the set if `previousMigratedVersion !== versionToMigrate`
2. Only perform this set inside the catch and add a condition on line 147 that only performs the set if `previousMigratedVersion !== versionToMigrate`

I prefer `1` because it's a little bit safer than `2`, specially on unhandled crashes (e.g.: OOM), but has a bit more I/O.

index.js
satazor: Done!

index.js
rafaelramalho19: I asked that on https://github.com/sindresorhus/conf/pull/83#discussion_r306029725

Let's just see what he says.

index.js
```diff
@@ -97,6 +103,16 @@ class Conf {
 		} catch (_) {
 			this.store = store;
 		}
+
+		const {migrations, projectVersion} = options;
+
+		if (migrations) {
+			if (!projectVersion) {
+				throw new Error('You need to specify the `projectVersion` in order for the migrations to work!');
```
satazor: I see that you changed this to an error which seems better. Perhaps make it a TypeError?

index.js
satazor: Actually I've gave a second though about this and perhaps logging is not a good idea. This library is being used by a lot of CLI tools and having control over logging is crucial. Having a library perform logging is not ideal, specially if we don't have any control over it.

Anyway, lets see what @sindresorhus has to say.

index.js
rafaelramalho19: I've left the `set` in the last line for the following case:

(Assume the last time we ran the migrations it was on version 1.0.0)
```js
const migrations = {
    '1.0.0': () => initialMigration,
    '1.0.1' : () => intermediateMigration,
    '1.0.2': () =>  lastMigration
}
```

and let's say we are on version `1.0.3`

If the next time we add a migration, e.g.: 
```js
const newMigrations = {
    '1.0.0': () => initialMigration,
    '1.0.1' : () => intermediateMigration,
    '1.0.2': () =>  lastMigration,
    '1.0.3': () =>  actuallyLastMigration,
}
```
(and the version is still 1.0.3), do we want the `actuallyLastMigration` to run? 

I assumed that we want the same as we discussed previously.

HOWEVER, one could argue that in this case, it makes sense. Because although we didn't change versions (1.0.3 -> 1.0.3), we still would want the migration code from a set of versions that weren't matched before (1.0.2 -> 1.0.3) 🤔

readme.md
```diff
@@ -97,6 +97,36 @@ config.set('foo', '1');
...
+### migrations
+
+Type: `Object`
+
+You can use migrations to perform operations to the store whenever a version is switched.
+
+The `migrations` object should be consisted of a key-value pair of `version`: `handler`.
+
+**Note**: The `projectVersion` option should be specified in order for the migrations to be run.
+But you only need to specify the version if you don't have a package.json file in your project.
```
satazor: Perhaps change this line to emphasize that we assume the value of the `version` key. The `version` field might not be set in the package.json.

> You only need to specify the `projectVersion` if there's no version defined in the project's package.json file or there's no package.json at all.

satazor: Also, I think it's worth documenting the `projectVersion` option, after the `projectName`  one.

index.js
rafaelramalho19: Do you have any recommendation/preference for the stubbing library?

index.js
rafaelramalho19: Done, please review again 😅

readme.md
rafaelramalho19: Agree! Nice catch.

index.js
```diff
@@ -51,7 +52,12 @@ class Conf {
...
-				options.projectName = pkgPath && JSON.parse(fs.readFileSync(pkgPath, 'utf8')).name;
+				const pkg = pkgPath && JSON.parse(fs.readFileSync(pkgPath, 'utf8'));
+
+				if (pkg) {
+					options.projectName = pkg.name;
+					options.projectVersion = pkg.version;
+				}
 			}
 
 			if (!options.projectName) {
```
satazor: I think we have an issue here: If I specify the `projectName` option manually, the `options.projectVersion` default value assignment based on the package.json version would never be performed.

This means that the code on line `59` (`options.projectVersion = pkg.version;`) needs to be run independently of having `options.projectName` set manually or not.

index.js
rafaelramalho19: Wanted to keep consistency with line 64
throw new Error('Project name could not be inferred. Please specify the `projectName` option.');

But I actually ended up not being consistent at all 😅 

Maybe I should update to:
throw new Error('Project version could not be inferred. Please specify the `projectVersion` option.');
?

index.js
satazor: I usually use Jest's built-in way of stubbing/mocking. But since we are not using Jest, we may use sinon or something. But I would wait for @sindresorhus feedback as it might not be necessary at all if we remove the logging.

readme.md
```diff
@@ -113,6 +143,13 @@ Default: The `name` field in the package.json closest to where `conf` is importe
 
 You only need to specify this if you don't have a package.json file in your project.
 
+#### projectVersion
+
+Type: `string`<br>
+Default: The `version` field in the package.json closest to where `conf` is imported.
+
+You only need to specify this if you don't have a package.json file in your project.
```
satazor: You only need to specify this if you don't have a package.json file in your project or if there isn't a version defined within it.

rafaelramalho19: Done. I did the same for the name!

index.js
satazor: I think that my suggestion is unrelated to what you described. The change I suggested is to avoid having to do `this.set(MIGRATION_KEY, versionToMigrate);` if the value is already that: `versionToMigrate === previousMigratedVersion`.

index.js
satazor: Agree.

readme.md
```diff
@@ -97,6 +97,36 @@ config.set('foo', '1');
...
+### migrations
+
+Type: `Object`
+
+You can use migrations to perform operations to the store whenever a version is switched.
+
+The `migrations` object should be consisted of a key-value pair of `version`: `handler`.
+
+**Note**: The `projectVersion` option should be specified in order for the migrations to be run.
+But you only need to specify the `projectVersion` if you don't have a package.json file in your project.
```
satazor: Perhaps remove this line which is similar to what is described in the option itself and link `projectVersion` to the option itself in the sentence above (The **`projectVersion`** option...)?

index.js
rafaelramalho19: You're right. I abstracted the `pkg` logic into a legible function and re-used it in 2 separate code blocks. 


I wanted to also add a test to cover this situation you described so it doesn't happen in the future. However, it seems that there are no actual tests for the inferring values from the package.json.

Maybe this is something I can update when we add a stubbing lib?

readme.md
rafaelramalho19: Done, seems better now?

index.js
satazor: Just something to have in consideration is that this will read the package.json 2x in the worst case scenario, which is when there’s no name nor version defined. The worst case scenario is the most common one, so perhaps we should optimize for it by reading the package data if either the name or version is missing, assign it to a const, and then use it as we are already doing.

index.js
rafaelramalho19: You're right, sorry about that.

Added the semver's comparision function, just in case,

`semver.eq(previousMigratedVersion, versionToMigrate)`.

rafaelramalho19: You're right, sorry about that.

Added the semver's comparision function, just in case,

`semver.eq(previousMigratedVersion, versionToMigrate)`.

test.js
```diff
@@ -290,7 +290,7 @@ test('safely handle missing package.json', t => {
 
 	let config;
 	t.notThrows(() => {
-		config = new Conf({projectName: 'conf-fixture-project-name'});
+		config = new Conf({projectName: 'conf-fixture-project-name', projectVersion: '1.0.0'});
```
satazor: I think this change is unnecessary since the migration option is not being specified.

index.js
satazor: Another remark: perhaps you can remove the if conditions when assigning options.projectName and options.projectVersion, and just assign them directly even if they are falsy.

rafaelramalho19: Yes, you're right. Done :)

I was afraid of committing the same error as in https://github.com/sindresorhus/conf/pull/49#discussion_r279627139, but it actually doesn't impact anything regarding this.

Thanks for catching that up.

test.js
rafaelramalho19: But it now breaks, as it's not "secured" under the protective code of `if (!options.projectName)` on line 53.

index.js
satazor: Actually, the problem still exists if there `options.cwd` was specified. We need to move it all way up to when the options and their defaults are being set.

Moreover,, you might want to avoid reading the package.json if not necessary (when cwd or projectName were specified AND projectVersion was specified as well). This is minor, but might be worth doing if the code remains simple.

test.js
satazor: Hmm, I referring to not having to specify the projectVersion in this line, effectively removing the change. Or maybe I’m missing something?

index.js
vasco-santos: I am totally in favor of having jsdocs. However, as adding jsdocs to all the other functions is not in the scope of this PR, I would like to propose a different approach:

1) Keep this jsdoc and create a new PR with jsdocs to all the remaining functions.
2) Create a new PR following this one with jsdocs for everything.

I prefer the first approach, but let's hear from @sindresorhus

test.js
vasco-santos: We should avoid breaking the current usage of the module.

Line 61 condition should only be true if we set the `migration` option. We can improve that conditional logic, in order to prevent this.

rafaelramalho19: You're both absolutely right (my brain was probably asleep last night).

I added a validation for the existance of  `packageData` (line 53 and 61), which was causing the compatibility to break.

Sorry about that 👍

index.js
```diff
@@ -47,11 +48,22 @@ class Conf {
 		};
 
 		if (!options.cwd) {
-			if (!options.projectName) {
-				const pkgPath = pkgUp.sync(parentDir);
-				// Can't use `require` because of Webpack being annoying:
-				// https://github.com/webpack/webpack/issues/196
-				options.projectName = pkgPath && JSON.parse(fs.readFileSync(pkgPath, 'utf8')).name;
+			const packageData = this._getPackageData();
```
rafaelramalho19: Nifty pick: I would probably like to refactor this to something like `_inferProjectDataFromPackage` to highlight the goal of this function. 

Thoughts on this?

readme.md
satazor: Yep, but I think you have missed documenting that handlers must be idempotent and briefly explain why (they will run more than once if migrations fail).

index.js
satazor: I agree. Let’s try it.

sindresorhus:
Doc comments should be in https://github.com/sindresorhus/conf/blob/master/index.d.ts, which you also need to update.

index.js
sindresorhus: I don't think we should catch migration errors. If migrations fail, the conf state will not be in the expected state and other things will probably randomly fail too.

sindresorhus:
Please see my feedback for the previous attempt at this in https://github.com/sindresorhus/conf/pull/49#issuecomment-487836538.

index.js
rafaelramalho19: On a closer look, the function doesn't actually infer the data from the package.json, it just provides it. 
So I guess I'll leave it as it is 🤷‍♂

index.js
```diff
@@ -47,11 +48,22 @@ class Conf {
 		};
 
 		if (!options.cwd) {
```
satazor: I've already commented it below, but if `options.cwd` is defined, the logic of getting the default `options.projectVersion` will never be run.  Worth adding a test for this so that you see that it's an issue with the current implementation.

index.js
```diff
@@ -47,11 +48,22 @@ class Conf {
...
-				const pkgPath = pkgUp.sync(parentDir);
-				// Can't use `require` because of Webpack being annoying:
-				// https://github.com/webpack/webpack/issues/196
-				options.projectName = pkgPath && JSON.parse(fs.readFileSync(pkgPath, 'utf8')).name;
+			const packageData = this._getPackageData();
+
+			if (packageData && !options.projectName) {
+				const {name} = packageData;
+
+				if (name) {
```
satazor: We cant simplify these conditions, we can just do:

```js
if (!options.projectName) {
   options.projectName = packageData.name;
}
```

which was like it was before.. and also:

if (!options.projectVersion) {
   options.projectVersion= packageData.version;
}

rafaelramalho19:
Hi @sindresorhus, regarding your comment on 

> Please see my feedback for the previous attempt at this in [#49 (comment)](https://github.com/sindresorhus/conf/pull/49#issuecomment-487836538).

Is it possible to use, for example, `store.set('__internal__.version', '1.2.3')` ?

index.js
```diff
@@ -112,6 +141,56 @@ class Conf {
...
+		const newerVersions = Object.keys(migrations)
+			.filter(candidateVersion => this._shouldPerformMigration(candidateVersion, previousMigratedVersion, versionToMigrate))
+			.sort(semver.compare);
+
+		newerVersions.forEach(version => {
+			try {
+				const migration = migrations[version];
+				migration(this);
+
+				this.set(MIGRATION_KEY, version);
```
satazor: About the current implementation and idem-potency: after reading how the store works, it's actually atomic per change!

This means that we can restore the state of the store after a migration error. Something like this:

```js
const storeBackup = Object.assign({}, this.store);  // Or use spread if we can depending on the nodejs version we support
    
try {
    const store = Object.assign({}, this.store);  // Or use spread if we can depending on the nodejs version we support
    
    const migration = migrations[version];
    migration(this);

    this.set(MIGRATION_KEY, version);
} catch (err) {
    this.store = storeBackup;
    throw err;
}
```

**Still**, if `this.store = storeBackup;` fails for some reason (e.g.: disk full), the handler will run more than once. @sindresorhus can you weight your opinion here?

sindresorhus:
> Is it possible to use, for example, store.set('__internal__.version', '1.2.3') ?

Do you mean for users? Then yes, but it should not be allowed. We should document that the `__internal__` key is reserved and throw if it's used from the outside. I doubt anyone would ever encounter it though.

index.js
satazor: I guess he's just catching to have a more friendly message and throwing the error again.

readme.md
rafaelramalho19: > > Is it possible to use, for example, store.set('**internal**.version', '1.2.3') ?
> 
> Do you mean for users? Then yes, but it should not be allowed. We should document that the `__internal__` key is reserved and throw if it's used from the outside. I doubt anyone would ever encounter it though.

I mean for internal usage. If the user has set `accessPropertiesByDotNotation` to false, the internal migration code would fail, right?

index.js
```diff
@@ -112,6 +141,56 @@ class Conf {
...
+			} catch (error) {
+				console.error('Something went wrong during the migration!');
+				console.error('Changes applied to the store until this point will be persisted.');
+				throw error;
+			}
+		});
+
+		this.set(MIGRATION_KEY, versionToMigrate);
```
satazor: This operation still executes 1 time unnecessary in the following scenario:

```
previousMigratedVersion = 1.0.1
versionToMigrate = 1.0.1
```

It will run `this.set(MIGRATION_KEY, 1.0.1);` inside the `forEach` loop and again here. To solve this, simply move the `semver.eq` check here and update `previousMigratedVersion` inside the loop

```js
               // inside the loop..
               this.set(MIGRATION_KEY, version);
               previousMigratedVersion = version;

        // outside the loop, at the bottom..
	if (!semver.eq(previousMigratedVersion, versionToMigrate)) {
		this.set(MIGRATION_KEY, versionToMigrate);
	}
```

rafaelramalho19:
> > Is it possible to use, for example, store.set('**internal**.version', '1.2.3') ?
> 
> Do you mean for users? Then yes, but it should not be allowed. We should document that the `__internal__` key is reserved and throw if it's used from the outside. I doubt anyone would ever encounter it though.

I mean for internal usage. If the user has set `accessPropertiesByDotNotation` to false, the internal migration code would fail, right?

readme.md
rafaelramalho19: Do you want to wait for the discussion in https://github.com/sindresorhus/conf/pull/83#discussion_r306199633 to proceed before I apply the change you suggested?

satazor:
@rafaelramalho19 worth closing all the conversations that were already taken care since there's a lot of conversations in the PR :D

index.js
rafaelramalho19: You're right, updated it and the tests!

Can you check again? :)

index.js
rafaelramalho19: I don't think we can, or am I missing something?

If the package.json couldn't be inferred (it happens in the tests for example) it causes the code to break (`cannot read property version of undefined`)

satazor: Sorry, I've updates the comment above.

index.js
rafaelramalho19: You're right since we're filtering the versions, the early exit is a kind of pointless. However, one could argue that it does prevent the filter from iterating the migrations object.

Updated it, thanks 👍

index.js
rafaelramalho19: Done 👍

index.js
```diff
@@ -46,12 +47,15 @@ class Conf {
 			...options
 		};
 
+		const packageData = this._getPackageData();
```
satazor: We may further simplify some more code by doing:

```js
               const packageData = this._getPackageData();

		options = {
			configName: 'config',
			fileExtension: 'json',
			projectSuffix: 'nodejs',
			clearInvalidConfig: true,
			serialize: value => JSON.stringify(value, null, '\t'),
			deserialize: JSON.parse,
			accessPropertiesByDotNotation: true,
                        projectName: packageData && packageData.name,
                        projectVersion: packageData && packageData.version,
			...options
		};
```
..and remove the assignment of these options from below.

Also note that having `this._getPackageData` is no longer necessary and we may inline it just like it was before. Just something to consider.

rafaelramalho19: Doesn't your suggestion always remove any specified `projectName` and `projectVersion` by the client?

At least for the name, it should only be inferred if the `cwd` option is not specified.

See https://github.com/sindresorhus/conf/commit/bb26c4f0a64cd8f1868eda2883cea651f0462fc0

index.js
rafaelramalho19: What @satazor said 😄 

If you're fine with the current solution, can I choose a stubbing library or do you prefer one?

If you prefer to just throw the error without a friendlier message, please disregard my last question.

readme.md
satazor: Lets wait for @sindresorhus feedback. Still, there are situations in which the function might run twice, even if we make the effort of prevent most of them. So the comment still makes sense to be added here.

index.js
satazor: My suggestion just set the defaults of `options`, which the user may override. This means that if the client supplies a custom `projectName` and/or `projectVersion`, those will take precedence because we are mixin in.

satazor: Checking https://github.com/sindresorhus/conf/commit/bb26c4f0a64cd8f1868eda2883cea651f0462fc0, the current code also re-introduces the bug because we are reading the package.json, even if we don't need to (both cwd and projectVersion were specified by the client).

To solve gracefully, ignore my comment above and perhaps do this instead:

```
    _fillOptionsFromPackageData(options) {
        // Avoid having to read package.json if not necessary
        // See: https://github.com/sindresorhus/conf/pull/62
        if (options.cwd && (!options.migrations || options.projectVersion)) {
              return;
        }

        const pkgPath = pkgUp.sync(parentDir);
	// Can't use `require` because of Webpack being annoying:
	// https://github.com/webpack/webpack/issues/196
        const pkgData = pkgPath && JSON.parse(fs.readFileSync(pkgPath, 'utf8'));

        if (!pkgData) {
             return;
        }

        if (!options.packageName) {
             options.packageName = pkgData.name;
        }

        if (!options.packageVersion) {
             options.packageVersion = pkgData.version;
        }
   }
```

and then use this function right after the `options` being assigned. You may then remove `_getPackageData` and all other unnecessary assignments that are already being done by this function.

satazor:
> > > Is it possible to use, for example, store.set('**internal**.version', '1.2.3') ?
> > 
> > 
> > Do you mean for users? Then yes, but it should not be allowed. We should document that the `__internal__` key is reserved and throw if it's used from the outside. I doubt anyone would ever encounter it though.
> 
> I mean for internal usage. If the user has set `accessPropertiesByDotNotation` to false, the internal migration code would fail, right?

I guess you have to create a internal `_set` function that doesn't care about key names, while the exposed pubic `set` ensures that keys can't start with **internal.**. Also, I would add more "uniqueness" to the key name, such as **__INTERNAL__.**

index.js
satazor: Another approach is to use something like this:

```js
// Place this inside the constructor 
const getPackageData = memoizeOne(() => {
        const pkgPath = pkgUp.sync(parentDir);
	// Can't use `require` because of Webpack being annoying:
	// https://github.com/webpack/webpack/issues/196
        const pkgData = pkgPath && JSON.parse(fs.readFileSync(pkgPath, 'utf8'));

       return pkgData || {};
});

// and then right below..

if (!options.cwd) {
   			if (!options.projectName) {
            			 options.projectName = getPackageData().name;
      			 }

        		// ...
}


if (migrations) {
   			if (!options.projectVersion) {
				options.projectVersion = getPackageData().version;
			}

			if (!options.projectVersion) {
				throw new Error('Project version could not be inferred. Please specify the `projectVersion` option.');
			}

			this._migrate(options.migrations, options.projectVersion);
		}
}
```

The advantage is that the code changes are minimal and most of the logic remains the same. `memoizeOne` refers to this: https://www.npmjs.com/package/memoize-one

Which one do you prefer?

satazor: Spoiler: I prefer the last one, seems more simple and the options logic is more centralized as they are handled just in a single place and consistently.

rafaelramalho19: I also prefer the memoization solution for the reason that it scales if you need more properties from the package.json.

And also prefer the code flow of the second option where the project version is only set when needed (if the migrations are being used).

rafaelramalho19: Done, thank you very much for the nice suggestions 😄

rafaelramalho19:
> > > > Is it possible to use, for example, store.set('**internal**.version', '1.2.3') ?
> > > 
> > > 
> > > Do you mean for users? Then yes, but it should not be allowed. We should document that the `__internal__` key is reserved and throw if it's used from the outside. I doubt anyone would ever encounter it though.
> > 
> > 
> > I mean for internal usage. If the user has set `accessPropertiesByDotNotation` to false, the internal migration code would fail, right?
> 
> I guess you have to create a internal `_set` function that doesn't care about key names, while the exposed pubic `set` ensures that keys can't start with `__internal__`.

Done!

Since we currently only need it for the migrations, I only used `dot notations` for now. (This skips the validation for `this._options.accessPropertiesByDotNotation` and all the validations associated with it.

If you prefer I can modify the `set` function flow to accept an override mechanism for the `accessPropertiesByDotNotation`.

index.js
satazor: @rafaelramalho19 bumping this thread to ensure you didn’t forgot to remove the jsdoc.

rafaelramalho19: I didn't, I'm just waiting on @sindresorhus answer 😄 

But I can remove it, for the time being, and add it on a later merge request.

index.js
sindresorhus: It should only run once. Double-migrations can cause unforeseen problems.

rafaelramalho19: Alright @sindresorhus, will update it 😄

index.js
sindresorhus: I missed that you were throwing the error too. I would prefer the best of both worlds and modify the original error with a more useful message, but also include the original message in the error. So no `console.error` logging.

index.js
```diff
@@ -112,6 +135,61 @@ class Conf {
...
+	_migrate(migrations, versionToMigrate) {
+		let previousMigratedVersion = this._get(MIGRATION_KEY, '0.0.0');
+
+		const newerVersions = Object.keys(migrations)
+			.filter(candidateVersion => this._shouldPerformMigration(candidateVersion, previousMigratedVersion, versionToMigrate))
+			.sort(semver.compare);
+
+		let storeBackup = {...this.store};
+
+		newerVersions.forEach(version => {
```
sindresorhus: Use a `for-of` loop. (Just my preference)

index.js
sindresorhus: Would be useful to also include which step in the migration that failed.

index.js
```diff
@@ -129,6 +207,10 @@ class Conf {
 			throw new TypeError('Use `delete()` to clear values');
 		}
 
+		if (typeof key === 'string' && key.indexOf(INTERNAL_KEY) === 0) {
```
sindresorhus: `key.indexOf(INTERNAL_KEY) === 0` will match even `__internal__foo` which is not something people would do, probably, but it's still an incorrect check.

index.js
```diff
@@ -46,12 +51,18 @@ class Conf {
 			...options
 		};
 
+		const getPackageData = memoizeOne(() => {
```
sindresorhus: You don't really need memoization here, you can just use https://github.com/sindresorhus/onetime

readme.md
```diff
@@ -97,6 +97,35 @@ config.set('foo', '1');
...
+            store.delete('debug phase');
+            store.set('phase', '1.0');
+       },
+       '1.0.2': store => {
+            store.set('phase', '>1.0');
+       }
+    }
+});
+```
+
```
sindresorhus: Same feedback as your electron-store PR.

readme.md
```diff
@@ -111,7 +140,14 @@ Useful if you need multiple config files for your app or module. For example, di
 Type: `string`<br>
 Default: The `name` field in the package.json closest to where `conf` is imported.
 
-You only need to specify this if you don't have a package.json file in your project.
+You only need to specify this if you don't have a package.json file in your project or if there isn't a name defined within it.
```
sindresorhus:
```suggestion
You only need to specify this if you don't have a package.json file in your project or if it doesn't have a `"name"` field.
```

index.js
sindresorhus: As for stubbing, https://sinonjs.org/ is good.

rafaelramalho19: Your suggestion makes total sense. Updated it 😄

readme.md
rafaelramalho19: Done 😄

index.js
rafaelramalho19: 👍

index.js
rafaelramalho19: Makes sense! Updated it and added 2 tests to check your case and when `accessPropertiesByDotNotation` is set to false

index.js
rafaelramalho19: Done :)

readme.md
rafaelramalho19: Done, also replaced it in other 3 places.

rafaelramalho19:
Hi @sindresorhus, can you provide any update on this PR? 😄

sindresorhus:
This is excellent. Thank you for working on this, @rafaelramalho19. And sorry for the delay. The summer got the best of me and I'm way behind on OSS work.

sindresorhus:
I noticed that there are no tests or docs (and TS) for using a semver range in the migration keys. Would you be able to submit a follow-up PR for that?

rafaelramalho19:
It's okay. I hope you had a nice summer @sindresorhus 😄 

I'll create a follow-up PR this week and tag you.

PR #87, opened 2019-09-09
Add `.reset()` method

husek:
Linked with: https://github.com/sindresorhus/electron-store/pull/84

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -268,7 +268,7 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	/**
 	 Reset an item to its default value, as defined on `defaults` or its `schema` option.
 
-	 @param keys - The key/keys of the item/items to reset.
+	 @param keys - The keys of the items to reset.
 	 */
 	reset<K extends keyof T>(...keys: K[]): void;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -266,6 +266,13 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	get<K extends keyof T>(key: K, defaultValue?: T[K]): T[K];
 
 	/**
+	 Reset an item to its default value (as defined on its schema).
+
+	 @param key - The key of the item to get.
+	 */
+	reset<K extends keyof T>(key: K): void;
+
+	/**
 	Check if an item exists.
 
 	@param key - The key of the item to check.
--- a/index.d.ts
+++ b/index.d.ts
@@ -266,11 +266,11 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	get<K extends keyof T>(key: K, defaultValue?: T[K]): T[K];
 
 	/**
-	 Reset an item to its default value (as defined on its schema).
+	 Reset an item to its default value, as defined on `defaults` or its `schema` option.
 
-	 @param key - The key of the item to get.
+	 @param keys - The key/keys of the item/items to reset.
 	 */
-	reset<K extends keyof T>(key: K): void;
+	reset<K extends keyof T>(...keys: K[]): void;
 
 	/**
 	Check if an item exists.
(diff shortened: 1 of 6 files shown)


sindresorhus:
It should also support resetting to the default value if the user defined the `defaults` option (which should also be documented).

index.d.ts
```diff
@@ -265,6 +265,13 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	*/
 	get<K extends keyof T>(key: K, defaultValue?: T[K]): T[K];
 
+	/**
+	Reset an item to its default value (as defined on its schema).
+
+	@param key - The key of the item to get.
```
sindresorhus: The description here is not correct.

sindresorhus:
I think it would be useful if it could reset multiple keys too.

index.test-d.ts
```diff
@@ -62,6 +63,8 @@ conf.set('unicorn', false);
 
 expectType<string>(conf.get('foo'));
 expectType<string>(conf.get('foo', 'bar'));
+conf.reset('foo');
+expectType<string>(conf.get('foo'));
```
sindresorhus: This assertion is meaningless, as it's just a type assertion and we already checked `.get`.

readme.md
```diff
@@ -279,6 +279,10 @@ Set multiple items at once.
 
 Get an item or `defaultValue` if the item does not exist.
 
+#### .reset(key)
+
+Reset an item to its predefined `defaultValue`.
```
sindresorhus: The description here should be the same as in the index.d.ts

test.js
```diff
@@ -11,6 +11,7 @@ const fixture = '🦄';
 
 test.beforeEach(t => {
 	t.context.config = new Conf({cwd: tempy.directory()});
+	t.context.configWithSchema = new Conf({cwd: tempy.directory(), schema: {foo: {default: 42}}});
```
sindresorhus: This should be defined in the test itself as it's only used once.

husek:
> It should also support resetting to the default value if the user defined the `defaults` option (which should also be documented).

Just added support to `defaults`

index.d.ts
```diff
@@ -265,6 +265,13 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	*/
 	get<K extends keyof T>(key: K, defaultValue?: T[K]): T[K];
 
+	/**
+	 Reset an item to its default value, as defined on `defaults` or its `schema` option.
+
+	 @param keys - The key/keys of the item/items to reset.
```
sindresorhus:
```suggestion
	 @param keys - The keys of the items to reset.
```

index.d.ts
```diff
@@ -265,6 +265,13 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	*/
 	get<K extends keyof T>(key: K, defaultValue?: T[K]): T[K];
 
+	/**
+	 Reset an item to its default value, as defined on `defaults` or its `schema` option.
```
sindresorhus:
```suggestion
	 Reset an item to its default value, as defined by the `defaults` or `schema` option.
```

index.js
```diff
@@ -266,6 +281,14 @@ class Conf {
 		return key in this.store;
 	}
 
+	reset(...keys) {
+		keys.forEach(key => {
```
sindresorhus: Use a `for-of` loop.

test.js
```diff
@@ -95,6 +95,23 @@ test('.has()', t => {
 	t.false(t.context.config.has('missing'));
 });
 
+test('.reset()', t => {
+	t.context.configWithSchema = new Conf({cwd: tempy.directory(), schema: {foo: {default: 42}, bar: {default: 99}}});
+	t.context.configWithDefaults = new Conf({cwd: tempy.directory(), defaults: {foo: 42, bar: 99}});
```
sindresorhus: Put the object items on separate lines. It's hard to read when everything is on one line.

index.d.ts
sindresorhus: It was implied that the readme text would also be updated.

PR #88, opened 2019-09-09
Add semver range capabilities into migrations

rafaelramalho19:
Related with https://github.com/sindresorhus/conf/pull/83#issuecomment-529325325

@sindresorhus I actually had to implement some code to make this work.
Can you please review it? 

I also included eslint configuration in this merge request, but I can split it into a separate one if you like. It was just making me angry not having the rules errors 😄


index.js
```diff
@@ -139,8 +139,7 @@ class Conf {
 		let previousMigratedVersion = this._get(MIGRATION_KEY, '0.0.0');
 
 		const newerVersions = Object.keys(migrations)
-			.filter(candidateVersion => this._shouldPerformMigration(candidateVersion, previousMigratedVersion, versionToMigrate))
-			.sort(semver.compare);
```
rafaelramalho19: The sort was breaking because semver can't sort this array: ['>1', '>=2'] 😢 

Any suggestions?

index.js
```diff
@@ -191,7 +190,19 @@ class Conf {
 		return false;
 	}
 
+	_isVersionInRangeFormat(version) {
```
sindresorhus: You should not have to check whether the version is in range format. Then the logic here is incorrect.

index.js
sindresorhus: https://github.com/npm/node-semver#comparison has a method to use for sorting semver.

Add some tests to confirm it's working.

rafaelramalho19: What method are you refering to? Is it the same that we were already using? `semver.compare`? 
Because it does not work with ranges

index.js
rafaelramalho19: `semver` requires different methods depending on whether we're using ranged or fixed versions.

`satisfies` only works with ranges, while `lte`, `gte`, etc, only works with fixed versions.
Do you suggest another approach to this? 😕

index.js
sindresorhus: If we can't sort them, we shouldn't really sort at all then. Why do we even need to sort? The object key order should be enough.

index.js
sindresorhus: It still feels weird having to use a regex to check for this. Seems like `semver` should solve that problem somehow. The regex is not correct either, it doesn't match versions like `1.1.1-beta`. You should probably use `semver.valid`.

index.js
rafaelramalho19: I guess so. 

However, sorting them would just allow the user to configure something like:

```js
const migrations = {
   '1.0.2': () => ....
   '0.0.1': () => ....
}
```

and it still would work in the correct order.


But it can be a non-issue, it's just harder and counter-intuitive to create migrations for older versions after a newer version migrate 🤔

index.js
rafaelramalho19: Yes, nice catch 👍

rafaelramalho19: Actually the most easy and straight-forward approach was to use `semver.clean` :)

Added a test to validate your use-case as well.

sindresorhus:
Thanks for implementing this. It would be great if you could take a look at https://github.com/sindresorhus/conf/issues/92 as it seems to be a pretty serious issue.

PR #89, opened 2019-09-10
Add support for setting/getting Date objects

sbencoding:
Implementing a new feature described in issue https://github.com/sindresorhus/electron-store/issues/18.  
Type information is stored under `${INTERNAL_KEY}.type` as suggested, the `get` method checks if there's a type key present, if there's it converts the value to the type before returning it.  
The `set` methods checks if the value passed in is an `instance of Date`, if it is it stores it as an object like:
```js
key: {
    __internal__.type: 'Date',
    val: valuePassedIn
}
````
I've also added a test to `test.js` to test for the setting and getting Date objects
Let me know if there's anything I could change. For example I chose to include type information only for Date objects, but it might be better to store type information for every key-value pair to be more consistent, but it would also make the object and the file larger.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -233,7 +233,6 @@ declare namespace Conf {
 		readonly watch?: boolean;
 
 		/**
-		Conf can automatically (de)serialize some objects/types.
 		As stated before `.set()` only accepts JSON serializable values, so you can't use `undefined`, `function` or `symbol`.
 		Additionally Conf has the ability to automatically serialize and deserialize pre-configured objects.
 		Currently supported objects are:
@@ -241,22 +240,22 @@ declare namespace Conf {
 		Extra types can be added to a Conf instance via passing the `extraTypes` option to the constructor.
 		Each element of the array defines an extra type, that is going to be automatically serialized when stored and deserialized when retrieved.
 
-		@example
+		@example ````The following example demonstrates the configuration of the `RegExp` type.
 		```
 		const Conf = new require('conf');
 
-		const extraTypeDate = {
-			name: 'Date',
-			isInstance: object => object instanceof Date,
-			convertFrom: value => value.getTime(),
-			convertTo: value = new Date(value)
+		const extraTypeRegExp = {
+			name: 'regex',
+			isInstance: object => object instanceof RegExp,
+			convertFrom: value => {pattern: value.source, flags: value.flags},
+			convertTo: value => new RegExp(value.pattern, value.flags)
 		};
-		const config = new Conf({extraTypes: [extraTypeDate]});
+		const config = new Conf({extraTypes: [extraTypeRegExp]});
 
-		config.set('myDate', new Date());
-		const myDate = config.get('myDate');
-		console.log(myDate.toDateString());
-		//=> Tue Apr 07 2020
+		config.set('myRegex', new RegExp('as.*', 'gi'));
+		const myRegex = config.get('myRegex');
+		console.log(myRegex.exec('asdfgh'));
+		//=> ['asdfgh']
 		```
 		*/
 		readonly extraTypes?: Conf.ExtraType<unknown>[];
@@ -272,14 +271,20 @@ declare namespace Conf {
 		name: string,
 		/**
 		Determines whether an object is of the defined type.
+
+		@returns True if the object is of the defined type, otherwise false.
 		*/
 		isInstance: (value: T) => boolean,
 		/**
 		Takes the object of the defined type and deserializes it.
+
+		@returns A JSON serializable representation of the custom type.
 		*/
 		convertFrom: (value: T) => JsonValue,
 		/**
 		Takes the serialized value and deserializes it.
+
+		@returns The type that is defined.
 		*/
 		convertTo: (value: JsonValue) => T
 	}
--- a/index.d.ts
+++ b/index.d.ts
@@ -238,13 +238,6 @@ declare namespace Conf {
 		readonly watch?: boolean;
 	}
 
-	interface ExtraType<T> {
-		name: string,
-		instance: T,
-		convertFrom: (val: T) => string | number,
-		convertTo: (val: string) => T
-	}
-
 	/**
 	 * Defines a type, that's (de)serialization is automatically handled
 	 */
@@ -264,7 +257,7 @@ declare namespace Conf {
 		/**
 		 * Convert the stored JsonValue to this type
 		 */
-		convertTo: (value: string) => T
+		convertTo: (value: JsonValue) => T
 	}
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -216,6 +216,13 @@ declare namespace Conf {
 		*/
 		readonly watch?: boolean;
 	}
+
+	interface ExtraType<T> {
+		name: string,
+		instance: T,
+		convertFrom: (val: T) => string | number,
+		convertTo: (val: string) => T
+	}
 }
 
 /**
@@ -225,6 +232,7 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	store: T;
 	readonly path: string;
 	readonly size: number;
+	readonly extraTypes: Conf.ExtraType<any>[];
 
 	/**
 	Changes are written to disk atomically, so if the process crashes during a write, it will not corrupt the existing config.
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,29 @@
 /// <reference types="node"/>
-import {JSONSchema} from 'json-schema-typed';
+import { JSONSchema } from 'json-schema-typed';
 
 declare namespace Conf {
 	type Schema = JSONSchema;
+
+	/**
+	 * Matches a JSON object
+	 */
+	type JsonObject = { [key: string]: JsonValue };
+
+	/**
+	 * Matches any valid JSON value
+	 */
+	type JsonValue = string | number | boolean | null | JsonObject | JsonArray;
+
+	/**
+	 * A list of currently supported extra types
+	 */
+	type SupportedExtraTypes = Date;
+
+	/**
+	 * Matches a JSON array
+	 */
+	interface JsonArray extends Array<JsonValue> { }
+
 
 	interface Options<T> {
 		/**
@@ -47,7 +68,7 @@ declare namespace Conf {
 
 		**Note:** The `default` value will be overwritten by the `defaults` option if set.
 		*/
-		readonly schema?: {[P in keyof T]: Schema};
+		readonly schema?: { [P in keyof T]: Schema };
 
 		/**
 		Name of the config file (without extension).
@@ -102,7 +123,7 @@ declare namespace Conf {
 		});
 		```
 		*/
-		readonly migrations?: {[version: string]: (store: Conf<T>) => void};
+		readonly migrations?: { [version: string]: (store: Conf<T>) => void };
 
 		/**
 		__You most likely don't need this. Please don't use it unless you really have to.__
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by sbencoding** [comment]
The travis CI tests/builds are passing, but node v12 on Windows failed to download node.js v12, so the code should be fine I think.

**Comment by sindresorhus** [comment]
This needs to be documented in the readme and index.d.ts. Also make sure that `$ npm test` passes locally.

File: index.d.ts [hunk 1]
```diff
@@ -202,6 +202,13 @@ declare namespace Conf {
 		*/
 		readonly accessPropertiesByDotNotation?: boolean;
 	}
+
+	interface ExtraType<T> {
```
sindresorhus: The interface and the individual properties need doc comments.

index.d.ts
```diff
@@ -202,6 +202,13 @@ declare namespace Conf {
...
+		name: string,
+		instance: T,
+		convertFrom: (val: T) => string | number,
```
sindresorhus: The return value here would be a good candidate for https://github.com/sindresorhus/type-fest/blob/eb110db80ccd74e8f5141cb52b3e99d3603399ce/source/basic.d.ts#L53

index.d.ts
```diff
@@ -248,7 +256,7 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	Set an item.
 
 	@param key - You can use [dot-notation](https://github.com/sindresorhus/dot-prop) in a key to access nested properties.
-	@param value - Must be JSON serializable. Trying to set the type `undefined`, `function`, or `symbol` will result in a `TypeError`.
+	@param value - Must be JSON serializable. Trying to set the type `undefined`, `function`, or `symbol` will result in a `TypeError`. Additionally Conf automatically seralizes and deserializes `Date` objects
```
sindresorhus: Typo

index.d.ts
```diff
@@ -211,6 +218,7 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	store: T;
 	readonly path: string;
 	readonly size: number;
+	readonly extraTypes: Conf.ExtraType<any>[];
```
sindresorhus: Should not be `any`.

index.js
```diff
@@ -214,6 +224,75 @@ class Conf {
...
+					[TYPE_KEY]: type.name,
+					val: type.convertFrom(value)
+				};
+			}
+		}
+
+		// Sub-object handling of extra types
+		const stack = [];
+		stack.push(value);
+		while (stack.length > 0) {
```
sindresorhus: Why are you using a while loop?

readme.md
```diff
@@ -321,6 +323,14 @@ conf.store = {
...
 Get the path to the config file.
 
+## Supported Types
+
+As stated before `.set()` only accepts JSON serializable values, so you can't use `undefined`, `function` or `symbol`.  
+Additionally Conf has the ability to automatically serialize and deserialize pre-configured objects.  
+Currently supported objects are:
+- `Date` - serializes the Date to milliseconds, then back to a `Date` object when the getter is called
+
+If you want support for a new object/type please open a new issue for discussion.
```
sindresorhus: The docs in the readme and index.d.ts should be in sync.

test.js
```diff
@@ -47,6 +47,18 @@ test('.set() - with object', t => {
...
+	t.context.config.set('dateobj', new Date());
+	const myDate = t.context.config.get('dateobj');
+	t.true(myDate instanceof Date);
+});
+
+test('.set() - with Date in a nested object', t => {
+	t.context.config.set('dateobj', {str: 'this is a string', date: new Date()});
+	const myDate = t.context.config.get('dateobj').date;
+	t.true(myDate instanceof Date);
+});
```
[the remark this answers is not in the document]
sbencoding: I've added more tests, but I'm not sure how to go about circular objects, becuase `Conf.serialize`, which is `JSON.stringify` by default, throws an exception is there's a circulation. So all I did for now is to detect if there's a circulation and stop handling that branch of the object, because it would cause an infinite loop, and cause the application to crash, but when `Conf` gets to serializing the object with `JSON` it still throws an error. I've also tested this behavior with `string` so it's not a problem with how the new feature handles extra types.

index.js
sbencoding: So in case of nested objects I want to reuse the same logic to serialize `Date` types, so there are two ways I can go about this. The first one is to use recursion, the other one is to use the iterative solution, which is the while loop.  
I've decided to use the iterative solution over the recursive solution, because with really deeply nested objects, there's a possibility that the function could overflow the callstack.

delewis13:
Any updates on merging this functionality? Would be great to have a simple way to store and retrieve dates.

sbencoding:
@delewis13 I've implemented the requested changes, but I didn't want to update the PR, because there are still questions, namely:
* Circular objects
In the review there's a request to handle circular objects, but even if I handle them `JSON.serialize()` just throws and exception when it detects one, so I'm not sure what to do here
* Usage of while loops
I've commented on the review explaining my thought process, but there might be other, better ways instead of using the while loop I haven't thought about.

Other than that the rest of the requested changes are implemented

papb:
Hi @sbencoding, there are other modules that can handle serialization and deserialization of circular objects automatically for you, such as [flatted](https://github.com/WebReflection/flatted), maybe this is what Sindre had in mind...

sbencoding:
Hey, sorry for the late response.
I've implemented the serialization and deserialization of `circular objects`, as well as I've added some tests for it. Other requested changes are implemented as well.
I'm not sure why the travis build fails tho....
* On all windows builds `xo` complains about import order in `test.js`, this doesn't happen for OSX or Linux or my local machine (Linux).
* OSX with nodejs v12 fails the both of the `watch` option tests. This doesn't happen for v10, and on the linux tests.

index.d.ts
```diff
@@ -1,9 +1,30 @@
 /// <reference types="node"/>
-import {JSONSchema} from 'json-schema-typed';
+import { JSONSchema } from 'json-schema-typed';
```
sindresorhus, line 2: Don't do unrelated style changes. Applies in many places. If you upgrade the XO dependency and run `npx xo --fix` these things will be fixed for you.

index.d.ts
```diff
@@ -1,9 +1,30 @@
...
 
 declare namespace Conf {
 	type Schema = JSONSchema;
 
+	/**
+	 * Matches a JSON object
+	 */
```
sindresorhus, line 9:
```suggestion
	/**
	Matches a JSON object.
	*/
```

Applies everywhere.

index.d.ts
```diff
@@ -225,6 +268,10 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 	store: T;
 	readonly path: string;
 	readonly size: number;
+	/**
+	 * Conf can automatically (de)serialize some objects/types, for more information please visit [Supported Types](https://https://github.com/sindresorhus/conf#Supported-Types)
```
sindresorhus, line 272: The readme and index.d.ts docs should be in sync. That means the linked to text should be included inline here.

index.js
```diff
@@ -283,6 +294,126 @@ class Conf {
...
+
+		// Sub-object handling of extra types
+		const stack = [];
+		const objectSeenList = [];
+
+		stack.push({current: value, parent: null, key: null});
+		let circularReferenceCounter = 1;
+		while (stack.length > 0) {
+			const {current, parent, key} = stack.pop();
+			if (objectSeenList.indexOf(current) === -1) {
```
sindresorhus, line 321:
```suggestion
			if (!objectSeenList.includes(current)) {
```

index.js
```diff
@@ -283,6 +294,126 @@ class Conf {
...
+					if (!didConvert) {
+						// Check sub-objects
+						stack.push({current: current[currentItemKey], parent: current, key: currentItemKey});
+					}
+				}
+			}
+		}
+
+		return value;
+	}
```
sindresorhus, line 366: This method is pretty large. I would look into splitting it into multiple methods and maybe something here is generic enough to be moved out into a generic utility function. Same with `_deserializeExtraTypes`.

index.js
```diff
@@ -283,6 +294,126 @@ class Conf {
...
+			const {current, parent, key} = stack.pop();
+			if (objectSeenList.indexOf(current) === -1) {
+				objectSeenList.push(current);
+			} else {
+				// Current object is reference to an object already parsed
+				// current is the circular reference
+				// seenList.indexOf(current) is the object it's pointing to
+				const referenceTarget = objectSeenList[objectSeenList.indexOf(current)];
+				parent[key] = {
+					[TYPE_KEY]: 'circularObject',
```
sindresorhus, line 329: I don't really see why you need the `circularObject` thing. Circular objects can simply be detected by comparing references.  I think this logic could be simplified a lot.

See how it's done here: https://github.com/sindresorhus/serialize-error/blob/master/index.js and here: https://github.com/sindresorhus/map-obj/blob/master/index.js

test.js
```diff
@@ -49,6 +49,76 @@ test('.set() - with object', t => {
...
+	const circular = {myDate: new Date(), c1: null, c2: null};
+	circular.c1 = circular;
+	circular.c2 = circular;
+	t.context.config.set('circular', circular);
+	const getCircular = t.context.config.get('circular');
+	t.true(getCircular.myDate instanceof Date);
+	t.true(getCircular.c1.myDate instanceof Date);
+	t.true(getCircular.c2.myDate instanceof Date);
+});
+
```
sindresorhus, line 121: There needs to be some tests for other types, just to ensure it works in general with custom types and other types than Date.

index.js
```diff
@@ -283,6 +294,126 @@ class Conf {
...
+					value: referenceTarget[CIRCULAR_REF_KEY] || circularReferenceCounter
+				};
+				if (referenceTarget[CIRCULAR_REF_KEY] === undefined) {
+					referenceTarget[CIRCULAR_REF_KEY] = circularReferenceCounter;
+					circularReferenceCounter++;
+				} else {
+					console.log(referenceTarget[CIRCULAR_REF_KEY]);
+				}
+
+				console.log('skipping pontentially circulation');
```
sindresorhus, line 339: Modules should not log to the console.

readme.md
```diff
@@ -356,6 +358,15 @@ conf.store = {
```
sindresorhus, line 358: You need to document how to use the custom deserializer/serializer.

index.js
sbencoding: Yeah, I'm very, very sorry about this, sometimes I just add some log statements for testing, and I always try to keep in mind to remove them, but it seems like I've forgotten this time.

index.js
sbencoding: The `circularObject` type signals to the deserialize function that it was a circular object, so it can lookup the `referenceNumber` set in the value of the object, and set it to the object it was referencing to before serialization.

I might have misunderstood this problem. I made it so that when deserializing the circulation is reconstructed, so you're going to end up with the same circular object after getting it from the store.
So should I just completely remove circulation like the `destroyCircular` function does in the first example you've provided, or is there an easier way to reconstruct the circulation?

sbencoding:
I've implemented most of the requested changes, I'm just going to re-work a bit the addition of extra types by the consumer program tomorrow, and then I'm going push the new changes.

I'm really confused about what I should do with circular objects now, but that's explained in a comment in the code review. I'd be happy to implement another approach, if I'm missing the point of this feature.

sbencoding:
I've implemented the requested changes, now the build only fails for node v12 on macOS during the 2 `.watch()` tests.

Instead of directly updating the extraTypes array of a Conf instance, I chose to implement the addition of extra types through the options argument that's passed to the constructor.

Let me know if there's anything else to change.

readme.md
```diff
@@ -356,6 +358,77 @@ conf.store = {
...
+
+Type: `Function`
+
+Takes the serialized value and deserializes it.
+
+- It has signature `convertTo(value)`, where value is a JSON serializable type.
+- It must return the type that is defined.
+
+#### Example
+The following example demonstrates the configuration of the `Date` type, that is internally supported by default.
```
sindresorhus, line 413: It would be better to use a different type than the default.

sindresorhus:
There are still some unresolved inline comments. Also make sure that the docs in the readme is in sync with index.d.ts and the other way.

sbencoding:
I've updated the example and double-checked to docs to reflect each other as much as possible, let me know if there's anything else to change/update.

sindresorhus:
I appreciate all the work done here, but I unfortunately don't think this is the optimal solution. Closing in favor of #154.

PR #90, opened 2019-09-10
Fix the TypeScript types

TheAifam5:
This PR fixes issues from the other PR I made:
https://github.com/sindresorhus/conf/pull/73

Regards,
TheAifam5

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -140,7 +140,7 @@ declare namespace Conf {
 
 		@default value => JSON.stringify(value, null, '\t')
 		*/
-		readonly serialize?: (value: Partial<T>) => string;
+		readonly serialize?: (value: T) => string;
 
 		/**
 		Function to deserialize the config object from a UTF-8 string when reading the config file.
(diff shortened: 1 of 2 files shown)


TheAifam5:
@jsynowiec If you want to use that right now, please type in your package.json following dependency:
```
"conf": "sindresorhus/conf#pull/90/head"
```

index.d.ts
```diff
@@ -255,7 +255,7 @@ declare class Conf<T = any> implements Iterable<[keyof T, T[keyof T]]> {
 
 	@param object - A hashmap of items to set at once.
 	*/
-	set(object: Partial<T>): void;
+	set(object: T): void;
```
sindresorhus: This has to be `Partial`, right? Since it should be possible to pass in only a subset of the types properties. Can you add a type test with a strongly-typed store for it?

jsynowiec: Yes, this should accept a subset, not the whole `T`

TheAifam5: Done!

sindresorhus:
Thanks everyone :)

sindresorhus:
Released: https://github.com/sindresorhus/conf/releases/tag/v6.0.1

If you use this for electron-store, just `rm -rf node_modules && npm i` to get it.

PR #91, opened 2019-09-14
Rewrite in TypeScript

### LINKED ISSUE #86: Improve the TypeScript types

	```
	index.test-d.ts:123:13
	```

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $80.00)

### Submitted pull Requests
---

### Tips

---
</details>

satyarohith:
closes #86

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,12 +13,14 @@
 		"node": ">=8"
 	},
 	"scripts": {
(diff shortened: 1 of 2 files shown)


satyarohith:
Will reopen if no one takes this in a month.

Issue #80, opened 2019-06-25
Feature Request: Allow schema to be free-form


d42ohpaz:
Currently the implementation expects that the schema provided is the `properties` of an object type. For most cases, this is probably fine, but I am finding it a little limiting because I am unable to define top-level properties, such as [`required`](https://json-schema.org/understanding-json-schema/reference/object.html#required-properties), [`propertyNames`](https://json-schema.org/understanding-json-schema/reference/object.html#property-names), [`additionalProperties`](https://json-schema.org/understanding-json-schema/reference/object.html#properties), et al. 

As such, I feel like projects could benefit from being able to define their own schema at any level, or default to the preexisting behavior: if the top level of the given schema doesn't define either `definitions` or `type`, then assume it's the original schema style and wrap it in its own `properties` key. Otherwise, allow the schema writer to define any level of the [JSON schema](https://json-schema.org/understanding-json-schema/reference/index.html). 

I'd be willing to work on this, but only if there is buy in.

sindresorhus:
Makes sense. I didn't think about that use-case. I just wanted to make it simple to define a schema without having to deal with the meta schema properties.

> if the top level of the given schema doesn't define either definitions or type, then assume it's the original schema style and wrap it in its own properties key.

This sounds too fragile. It's very likely someone would have a `type` key for a different purpose.

---

How do you suggest we add support for this? Either a breaking change or add a separate key for top-level meta properties.

9ae8sdf76:
> How do you suggest we add support for this? Either a breaking change or add a separate key for top-level meta properties.

I agree with your comment about guessing the schema form based on existence of a key. How about an alternative behavior:

By default treat all schemas as they currently are treated (i.e., properties of an existing object). Allow users to pass a flag with their options to opt-in their schema as free-form. No guesswork and backward compatible.

If you still don't like that, I think adding top-level keys for the meta properties would be just as good as anything else.

sindresorhus:
> Allow users to pass a flag with their options to opt-in their schema as free-form.

Can you show an example of how that would look like?

9ae8sdf76:
I was imaging it to look like:

```js
const Conf = require('conf');

const config = new Conf({implicitObjectRoot: false});

config.set({
    type: 'object',
    additionalProperties: {type: "string"},
    properties: {unicorn: '🦄'},
});
console.log(config.get('unicorn'));
//=> '🦄'

config.set('number', 1234); // causes validation error
[...]

PR #94, opened 2019-11-13
New option suggestion: direct assignment of ajv validate function

hokiedsp:
While the current support for JSON scheme validation is great, it limits more advanced usage of ajv. To fully unleash ajv capability, I'd love for conf to support an option to let user provide the ready-to-go ajv validator function rather than using the internally generated one via `schema` option. This PR implements such option, which I tentatively named `ajvValidator`.

Additional Notes on PR:
- `schema` option take precedence over `ajvValidator` option
- `ajvValidator` option is checked to be sure it's a function
- As far as I can tell, I don't think this PR affects the operation of `_validate()` unless user specifies a non-ajv validation function.
- I've added a test for the new option in `test.js` based on the existing `schema` option test

### DIFF
--- a/index.js
+++ b/index.js
@@ -99,6 +99,12 @@ class Conf {
 					this._defaultValues[key] = value.default;
 				}
 			}
+		} else if (options.ajvValidator) {
+			if (typeof options.ajvValidator !== 'function') {
+				throw new TypeError('The `schema` option must be an object.');
+			}
+
+			this._validator = options.ajvValidator;
 		}
 
 		if (options.defaults) {
--- a/test.js
+++ b/test.js
@@ -696,6 +696,36 @@ test('schema - validate Conf default', t => {
 	}, 'Config schema violation: `foo` should be string');
 });
 
+test('ajvValidator - validate Conf default', t => {
+	const Ajv = require('ajv');
+	const ajv = new Ajv({
+		allErrors: true,
+		format: 'full',
+		useDefaults: true,
+		errorDataPath: 'property'
+	});
+	const schema = {
+		type: 'object',
+		properties: {
+			foo: {
+				type: 'string'
+			}
+		}
+	};
+
+	const validator = ajv.compile(schema);
+
+	t.throws(() => {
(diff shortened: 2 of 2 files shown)


sindresorhus:
Could you elaborate on your use-case for needing this? It might help inform us on how we can better improve the defaults too.

hokiedsp:
@sindresorhus - Sure, the reason I added that option to my fork is to allow me to use the full schema files, specifically, to support $ref and use of multiple schema files. See [here for synchronous approach](https://github.com/epoberezkin/ajv#ref) or [here for async approach](https://github.com/epoberezkin/ajv#asynchronous-schema-compilation). 

> how we can better improve the defaults

I'm not sure if there is much room left to improve with the `schema` option as it provides only the `properties` property of a full JSON schema object. Expanding it in any way will disrupt its original behavior IMO.

My PR is the most generic one such that I can use ajv's async compilation with, but with your support I have an alternate ideas that I can implement for conf:

- `schemaObjects` option to accept an array of full schema objects, and use the first object's URI as the base
- `schemaFiles` option to accept an array of schema JSON file paths, read files, and use the `$id` of the first file as the base URI

These are more complete than the PR's `ajvValidator` but obviously are more complex and more work. As a side note, `schemaValidator` might be a better name for `ajvValidator`.

hokiedsp:
@sindresorhus - I just realized that you have an ongoing discussion and reached more-or-so the exact same conclusion: https://github.com/sindresorhus/conf/issues/80#issuecomment-529536104

As stated above, I'd push for `schemaObjects`/`rootSchema` to accept an array of JSON schema objects to support schema spanning multiple schema objects.

sindresorhus:
Too be honest, I'm not that well versed in JSON schema. I just added it because there was strong demand. It's not clear to me why you need support for multiple schema objects? It kinda seems to me that to solve all use-cases, there should be a `schemaMeta` option that accepts both `ajv` options add schemas. I don't really want to introduce more than one top-level option for this, so we need to figure out a good way to design it.

sindresorhus:
Closing for lack of response.

PR #95, opened 2019-11-18
Improve migration docs

### LINKED ISSUE #92: Migration creates wrong internal version number
The following code should create an config.json file with the config version number "0.0.1", instead it stores an internal version of "1.0.0":

```js
const Conf = require('conf')

new Conf({
  migrations: {
    "0.0.1": store => {
      store.set('a', 'foo');
    }
  }
});
```

config.json:
```json
{
	"a": "foo",
	"__internal__": {
		"migrations": {
			"version": "1.0.0"
		}
	}
}
```

rafaelramalho19:
Fixes #92

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -103,7 +103,7 @@ Type: `object`
 
 **Don't use this feature until [this issue](https://github.com/sindresorhus/conf/issues/92) has been fixed.**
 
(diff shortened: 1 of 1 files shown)


sindresorhus:
Thanks :)

PR #96, opened 2019-11-19
Rewrite in TypeScript

### LINKED ISSUE #86: Improve the TypeScript types

rafaelramalho19:
Closes #86

It's my first time working on Typescript, so this can be a bit bad.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#86: Improve the TypeScript types](https://issuehunt.io/repos/61670221/issues/86)
---

</details>


.eslintrc
```diff
@@ -0,0 +1,4 @@
+{
+    "extends": ["xo", "xo-typescript"],
+    "parser": "@typescript-eslint/parser"
+}
```
satazor: Missing new line

.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 yarn.lock
+.nyc_output
+dist
```
satazor: Same

tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+		"outDir": "dist",
+		"target": "es2017",
+		"lib": [
+			"es2017"
+		]
+	},
+	"include": [
+		"source/*"
+	]
+}
```
satazor: Same

source/index.ts
```diff
@@ -1,28 +1,26 @@
...
+import EventEmitter = require('events');
+import dotProp = require('dot-prop');
+import makeDir = require('make-dir');
+import pkgUp = require('pkg-up');
+import envPaths = require('env-paths');
+import writeFileAtomic = require('write-file-atomic');
+import Ajv = require('ajv');
+import debounceFn = require('debounce-fn');
+import semver = require('semver');
+import onetime = require('onetime');
```
papb: Is this syntax correct? Shouldn't it be `import fs from 'fs'` and the like?

rafaelramalho19: Although it does work, it's not the correct way.

Thanks for noticing this 😄

sindresorhus: No, this syntax is actually correct.

source/index.ts
```diff
@@ -39,21 +37,67 @@ const checkValueType = (key, value) => {
...
+export default class Conf {
+	_options: ConfOptions;
+
+	_defaultValues: ConfDefaultValues = {};
+
+	_validator?: Ajv.ValidateFunction;
+
+	encryptionKey?: string;
+
+	events?: EventEmitter;
```
sindresorhus: Why is this optional? Same with some of the others here.

sindresorhus:
- The doc comments in index.d.ts should be moved to the new index.ts.
- Try to make sure that the output `dist/index.d.ts` matches the old `index.d.ts` as close as possible.
- Don't use `any` or `object`. Use a specific type whenever possible, and if it's supposed to be non-specific, use `unknown`.
- https://github.com/sindresorhus/typescript-definition-style-guide has a lot of useful info about best practices.

source/index.ts
sindresorhus: 2

source/index.ts
```diff
@@ -367,65 +416,74 @@ class Conf {
...
+				const password = crypto.pbkdf2Sync(this.encryptionKey, initializationVector.toString(), 10000, 32, 'sha512');
+				const decipher = crypto.createDecipheriv(encryptionAlgorithm, password, initializationVector);
+				return Buffer.concat([decipher.update(data.slice(17)), decipher.final()]);
 			}
 
-			data = this.deserialize(data);
+			// Legacy decryption without initialization vector
+			// tslint:disable-next-line
+			// eslint-disable-next-line node/no-deprecated-api
+			const decipher = crypto.createDecipher(encryptionAlgorithm, this.encryptionKey);
```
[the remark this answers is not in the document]
sindresorhus: Doesn't matter. It's just a legacy fallback.

source/index.ts
rafaelramalho19: Updated 👍

source/index.ts
```diff
@@ -367,65 +416,74 @@ class Conf {
...
-		this.events.on('change', onChange);
-		return () => this.events.removeListener('change', onChange);
+		this.events?.on('change', onChange);
+		return () => this.events?.removeListener('change', onChange);
 	}
 
-	get size() {
-		return Object.keys(this.store).length;
-	}
+	encryptData(data: any): Buffer | undefined {
```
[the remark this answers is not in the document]
rafaelramalho19: It actually needs to be

`encryptData(data: string | Buffer): string | Buffer`

Updated it!

package.json
```diff
@@ -51,14 +81,24 @@
 		"write-file-atomic": "^3.0.0"
 	},
 	"devDependencies": {
-		"@types/node": "^12.7.4",
+		"@sindresorhus/tsconfig": "^0.4.0",
+		"@types/node": "^12.12.9",
+		"@types/semver": "^6.2.0",
+		"@types/write-file-atomic": "^2.1.2",
+		"@typescript-eslint/eslint-plugin": "^1.11.0",
+		"@typescript-eslint/parser": "^1.11.0",
```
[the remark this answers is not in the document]
rafaelramalho19: Done :)

test/index.ts
```diff
@@ -1,36 +1,38 @@
...
+import clearModule = require('clear-module');
+import pEvent = require('p-event');
+import delay = require('delay');
+import test from 'ava';
+import Conf from '../source';
 
 const fixture = '🦄';
 
-test.beforeEach(t => {
+test.beforeEach((t: any) => {
```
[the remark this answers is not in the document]
rafaelramalho19: I've tried other solutions. But it seems to all throw errors when using `t.context.config<...>`

Any ideas?

source/index.ts
```diff
@@ -39,21 +37,67 @@ const checkValueType = (key, value) => {
...
-class Conf {
-	constructor(options) {
+type ConfSerializer = (args: any) => ArrayBuffer | Buffer | string;
+type ConfDeserializer = (args: any) => any;
+type ConfDefaultValues = {
+	[key: string]: object;
+};
+type ConfMigrations = {
+	[key: string]: (store: Conf) => void;
+};
```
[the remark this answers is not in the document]
rafaelramalho19: Done :)

test/index.ts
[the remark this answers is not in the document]
sindresorhus: https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md#typing-tcontext

source/index.ts
papb: Oh, I finally [figured it out](https://github.com/sindresorhus/mem/issues/31#issuecomment-475858825), amazing...

source/index.ts
```diff
@@ -224,7 +303,7 @@ class Conf {
 		}
 	}
 
-	_containsReservedKey(key) {
+	_containsReservedKey(key: string | object): boolean {
```
sindresorhus:
```suggestion
	_containsReservedKey(key: string | {[key: string]: unknown}): boolean {
```

rafaelramalho19:
> * The doc comments in index.d.ts should be moved to the new index.ts.
> * Try to make sure that the output `dist/index.d.ts` matches the old `index.d.ts` as close as possible.
> * Don't use `any` or `object`. Use a specific type whenever possible, and if it's supposed to be non-specific, use `unknown`.
> * https://github.com/sindresorhus/typescript-definition-style-guide has a lot of useful info about best practices.
> * Revert: [ed72882#diff-21e9ddd93162651bd36f6e5bbfca8460L1-R14](https://github.com/sindresorhus/conf/commit/ed72882797aac361f35d0c3f567d95d246740b26#diff-21e9ddd93162651bd36f6e5bbfca8460L1-R14)

Regarding:
1️⃣ I also moved the exports to the bottom of the file so it doesn't get in the way of reading the main part of the source code. Is that okay with you?
2️⃣ Done.
3️⃣ I've removed all I could. The ones that remained seem a bit necessary 🤔 
4️⃣ Thanks for the read 👍 
5️⃣ Done, I had to do that with the `import * from x` syntax since it was throwing deprecated warnings.

source/index.ts
rafaelramalho19: Done it manually, had to update other places :)

test/index.ts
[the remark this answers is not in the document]
rafaelramalho19: Thanks 👍

source/index.ts
rafaelramalho19: The only one was actually `_validator` 😅 

Nice catch

package.json
```diff
@@ -38,27 +42,63 @@
...
 		"make-dir": "^3.0.0",
 		"onetime": "^5.1.0",
 		"pkg-up": "^3.0.1",
 		"semver": "^6.2.0",
-		"write-file-atomic": "^3.0.0"
+		"write-file-atomic": "^3.0.1"
 	},
 	"devDependencies": {
-		"@types/node": "^12.7.4",
+		"@sindresorhus/tsconfig": "^0.4.0",
```
sindresorhus: Upgrade this to latest version.

package.json
```diff
@@ -38,27 +42,63 @@
...
+		"@types/node": "^12.12.11",
+		"@types/semver": "^6.2.0",
+		"@types/write-file-atomic": "^2.1.2",
+		"@typescript-eslint/eslint-plugin": "^1.13.0",
```
sindresorhus: Upgrade this to latest version.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+import makeDir = require('make-dir');
+import pkgUp = require('pkg-up');
+import envPaths = require('env-paths');
+import writeFileAtomic = require('write-file-atomic');
+import Ajv = require('ajv');
+import debounceFn = require('debounce-fn');
+import semver = require('semver');
+import onetime from 'onetime';
+
+const plainObject: () => object = () => Object.create(null);
```
sindresorhus: Typo

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+
+	if (nonJsonTypes.includes(type)) {
+		throw new TypeError(`Setting a value of type \`${type}\` for key \`${key}\` is not allowed as it's not supported by JSON`);
+	}
+};
+
+const INTERNAL_KEY = '__internal__';
+const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
+
+export default class Conf<T = any> implements Iterable<[string, T]> {
```
sindresorhus:
```suggestion
export default class Conf<T = unknown> implements Iterable<[string, T]> {
```

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+import EventEmitter = require('events');
+import dotProp = require('dot-prop');
+import makeDir = require('make-dir');
+import pkgUp = require('pkg-up');
+import envPaths = require('env-paths');
+import writeFileAtomic = require('write-file-atomic');
+import Ajv = require('ajv');
+import debounceFn = require('debounce-fn');
+import semver = require('semver');
+import onetime from 'onetime';
```
sindresorhus:
```suggestion
import onetime = require('onetime');
```

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+
+const INTERNAL_KEY = '__internal__';
+const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
+
+export default class Conf<T = any> implements Iterable<[string, T]> {
+	_options: Options;
+
+	_defaultValues: DefaultValues = {};
+
+	_validator?: Ajv.ValidateFunction;
```
sindresorhus: This should not be optional.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+
+	* [Symbol.iterator](): IterableIterator<[string, any]> {
+		for (const [key, value] of Object.entries(this.store)) {
+			yield [key, value];
+		}
+	}
+}
+
+export type Serializer = (...args: unknown[]) => string;
+export type Deserializer = (arg: string | Buffer) => unknown;
```
sindresorhus:
```suggestion
export type Deserializer = (data: string) => unknown;
```

It should not accept `Buffer`.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+	}
+
+	* [Symbol.iterator](): IterableIterator<[string, any]> {
+		for (const [key, value] of Object.entries(this.store)) {
+			yield [key, value];
+		}
+	}
+}
+
+export type Serializer = (...args: unknown[]) => string;
```
sindresorhus: It should not accept spreaded arguments. Please make sure to compare against the existing type definition.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+export type Deserializer = (arg: string | Buffer) => unknown;
+export type DefaultValues = {
+	[key: string]: object;
```
sindresorhus: Use something more specific than `object`.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+};
+export type Migrations = {
+	[key: string]: (store: Conf) => void;
+};
+
+export type Schema = object | boolean;
```
sindresorhus: This type is not correct. Look at the existing type definition.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+
+		The only use-case I can think of is having the config located in the app directory or on some external storage. Default: System default user [config directory](https://github.com/sindresorhus/env-paths#pathsconfig).
+	*/
+	cwd?: string;
+
+	/**
+		Config used if there are no existing config.
+		** Note: The values in `defaults` will overwrite the `default` key in the `schema` option.
+	*/
+	defaults?: Readonly<object>;
```
sindresorhus: Use something more specific than `object`.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+			...partialOptions
+		};
+
+		const getPackageData = onetime(() => {
+			const packagePath = pkgUp.sync({cwd: parentDir});
+			// Can't use `require` because of Webpack being annoying:
+			// https://github.com/webpack/webpack/issues/196
+			const packageData = packagePath && JSON.parse(fs.readFileSync(packagePath, 'utf8'));
+
+			return packageData || {};
```
sindresorhus:
```suggestion
			return packageData ?? {};
```

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+
+		const fileExtension = options.fileExtension ? `.${options.fileExtension}` : '';
+		this.path = path.resolve(options.cwd, `${options.configName}${fileExtension}`);
+
+		const fileStore = this.store;
+		const store = Object.assign(plainObject(), options.defaults, fileStore);
+		this._validate(store);
+		try {
+			assert.deepEqual(fileStore, store);
+		} catch (_) {
```
sindresorhus:
```suggestion
		} catch {
```

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+		}
+
+		if (semver.gt(candidateVersion, versionToMigrate)) {
+			return false;
+		}
+
+		return true;
+	}
+
+	_get(key: string, defaultValue?: unknown): any {
```
sindresorhus: Don't use `any`.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+		this.store = store;
+	}
+
+	/**
+	Get an item.
+
+	@param key - The key of the item to get.
+	@param defaultValue - The default value if the item does not exist.
+	*/
+	get(key: string, defaultValue?: unknown): unknown {
```
sindresorhus: The types here are not correct. It should be strongly typed. Please see existing type definition.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+		return key in this.store ? this.store[key] : defaultValue;
+	}
+
+	/**
+	Set an item.
+
+	@param key - You can use [dot-notation](https://github.com/sindresorhus/dot-prop) in a key to access nested properties.
+	@param value - Must be JSON serializable. Trying to set the type `undefined`, `function`, or `symbol` will result in a `TypeError`.
+	*/
+	set(key: string | {[key: string]: unknown}, value?: unknown): void {
```
sindresorhus: Same as above.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+
+		return key in this.store;
+	}
+
+	/**
+	Reset items to their default values, as defined by the `defaults` or `schema` option.
+
+	@param keys - The keys of the items to reset.
+	*/
+	reset(...keys: string[]): void {
```
sindresorhus: Same as above.

source/index.ts
```diff
@@ -0,0 +1,746 @@
...
+				currentValue = newValue;
+				callback.call(this, newValue, oldValue);
+			}
+		};
+
+		this.events?.on('change', onChange);
+		return () => this.events?.removeListener('change', onChange);
+	}
+
+	encryptData(data: string | Buffer): string | Buffer {
```
sindresorhus:
```suggestion
	_encryptData(data: string | Buffer): string | Buffer {
```

test/index.ts
```diff
@@ -808,22 +814,24 @@ test('`watch` option watches for config file changes by file write', async t =>
...
 	t.is(conf._get('__internal__.migrations.version'), '0.0.2');
 });
 
-test('migrations - should save the project version when a migration occurs', t => {
+test('migrations - should save the project version when a migration occurs', (t: any) => {
 	const cwd = tempy.directory();
 
 	const migrations = {
-		'0.0.3': store => {
+		'0.0.3': (store: any) => {
```
sindresorhus: Don't use `any`.

sindresorhus:
> Try to make sure that the output dist/index.d.ts matches the old index.d.ts as close as possible.

I don't think you have done this thoroughly enough. The types in the existing type definition are better in many places.

rafaelramalho19:
> > Try to make sure that the output dist/index.d.ts matches the old index.d.ts as close as possible.
> 
> I don't think you have done this thoroughly enough. The types in the existing type definition are better in many places.

You're right Sindre. Sorry about that.

I've been a bit busy lately but don't worry about this becoming stale, I'll just pick it up as soon as I can.

source/index.ts
rafaelramalho19: I don't see any typo in this line Sindre. What did you mean?

source/index.ts
rafaelramalho19: What typing do you suggest here? Because `string` is already equal to `K` (which is just the key of Conf)

source/index.ts
sindresorhus: No idea. Must have been a mistake on my part.

package.json
```diff
@@ -13,12 +14,14 @@
 		"node": ">=8"
 	},
 	"scripts": {
-		"test": "xo && ava && tsd"
+		"build": "del dist && tsc",
+		"test": "npm run build && xo source/** && nyc ava",
```
sindresorhus: Use the XO `files` config instead of specifying it here.

source/index.ts
```diff
@@ -0,0 +1,744 @@
...
+		this.store = store;
+	}
+
+	/**
+	Get an item.
+
+	@param key - The key of the item to get.
+	@param defaultValue - The default value if the item does not exist.
+	*/
+	get(key: string, defaultValue?: T): T | undefined {
```
sindresorhus: This should have an overload that makes the return value `T` when `defaultValue` is given.

So the definitions should be:

```ts
get(key: string): T | undefined {
get(key: string, defaultValue: T): T {
```

source/index.ts
sindresorhus: The answer is in the existing TS definition:

https://github.com/sindresorhus/conf/blob/97bcc918a109ba964f7ef8a091cc5a762ae03576/index.d.ts#L289

Yes, both resolve to a string, but having it strongly typed will let editors auto-complete the key names and let TS enforce only valid key names.

source/index.ts
```diff
@@ -0,0 +1,744 @@
...
+
+	* [Symbol.iterator](): IterableIterator<[string, T]> {
+		for (const [key, value] of Object.entries(this.store)) {
+			yield [key, value];
+		}
+	}
+}
+
+export type Migrations<T> = {
+	[key: string]: (store: Conf<T>) => void;
```
sindresorhus:
```suggestion
	[semverSpecifier: string]: (store: Conf<T>) => void;
```

source/index.ts
```diff
@@ -0,0 +1,744 @@
...
+		this.store = this.createPlainObject();
+	}
+
+	/**
+	Watches the given `key`, calling `callback` on any changes.
+
+	@param key - The key wo watch.
+	@param callback - A callback function that is called on any changes. When a `key` is first set `oldValue` will be `undefined`, and when a key is deleted `newValue` will be `undefined`.
+	*/
+	onDidChange(key: string, callback: (...args: unknown[]) => void): () => unknown {
```
sindresorhus: This should be strongly-typed.

https://github.com/sindresorhus/conf/blob/97bcc918a109ba964f7ef8a091cc5a762ae03576/index.d.ts#L316-L319

source/index.ts
```diff
@@ -0,0 +1,744 @@
...
+			delete store[key];
+		}
+
+		this.store = store;
+	}
+
+	/**
+	Delete all items.
+	*/
+	clear(): void{
```
sindresorhus:
```suggestion
	clear(): void {
```

source/index.ts
```diff
@@ -0,0 +1,744 @@
...
+			return data;
+		}
+
+		try {
+			// Check if an initialization vector has been used to encrypt the data
+			if (data.slice(16, 17).toString() === ':') {
+				const initializationVector = data.slice(0, 16);
+				const password = crypto.pbkdf2Sync(this.encryptionKey, initializationVector.toString(), 10000, 32, 'sha512');
+				const decipher = crypto.createDecipheriv(encryptionAlgorithm, password, initializationVector);
+				const slicedData: any = data.slice(17);
```
sindresorhus: Don't use `any`.

source/index.ts
```diff
@@ -0,0 +1,744 @@
...
+	}
+
+	get size(): number {
+		return Object.keys(this.store).length;
+	}
+
+	get store(): {[key: string]: T} {
+		try {
+			const data: string | Buffer = fs.readFileSync(this.path, this.encryptionKey ? null : 'utf8');
+			const dataString = this._encryptData(data) as string;
```
sindresorhus: You should not use `as string` here as encryptData could return a Buffer. Generally, stay away from `as` unless necessary as it could easily hide type errors.

sindresorhus:
There were still some unresolved inline comments like https://github.com/sindresorhus/conf/pull/96#discussion_r350161217

source/index.ts
```diff
@@ -0,0 +1,744 @@
...
+		fs.watch(this.path, {persistent: false}, debounceFn(() => {
+			// On Linux and Windows, writing to the config file emits a `rename` event, so we skip checking the event type.
+			this.events?.emit('change');
+		}, {wait: 100}));
+	}
+
+	_migrate(migrations: Migrations<T>, versionToMigrate: string): void {
+		let previousMigratedVersion: string = this._get(MIGRATION_KEY, '0.0.0') as string;
+
+		const newerVersions: string[] = Object.keys(migrations)
```
sindresorhus:
```suggestion
		const newerVersions = Object.keys(migrations)
```

package.json
rafaelramalho19: From what I read and tested, the `files` is only used in the `overrides` option. I don't want a override here, I just want the linting to run in the source folder.

Since there's no `includes` or `cwd` in `xo`, I see 2 options:
1. Ignore all folders except `source` with the `ignores` option.
2. Leave this part as it is.

Do you have any other suggestion about this?

source/index.ts
rafaelramalho19: Are you sure about the return type of the second overload?
There's a wide range of inputs to the function where it could return undefined.

one example, assuming the store has: 
`[cool-key]: undefined`

using `get('cool-key', 'this should be the fallback`)` will return `undefined`.

Is also raises the question:
- Do we want undefined, but declared, values in the store to be returned or fallbacked?

package.json
sindresorhus: Ah, right. Ok, let's go with 2 then.

source/index.ts
sindresorhus: The store cannot have `undefined`:

>The value must be JSON serializable. Trying to set the type undefined, function, or symbol will result in a TypeError.

My suggestion assumes you also do https://github.com/sindresorhus/conf/pull/96#discussion_r362317785 to all the set/get/reset/etc methods, so they're strongly typed.

sindresorhus:
@rafaelramalho19 Still interested in finishing this?

rafaelramalho19:
> @rafaelramalho19 Still interested in finishing this?

Yes, I was just a bit busier these few last weeks 😅

source/index.ts
rafaelramalho19: In the codebase the validator only exists if there's a schema. Doesn't that make it optional?

source/index.ts
rafaelramalho19: I've not been able to make it work, it always says this:

image

Do you have any suggestions for my problem? :/

rafaelramalho19:
> @rafaelramalho19 Still interested in finishing this?

Maybe someone else can pick up on this? I'm more than happy to let someone else finish this, I guess I'm not experienced enough in Typescript for converting all these code.

Sorry for the long wait Sindre.

sindresorhus:
Sure, no worries at all. I totally understand. TypeScript is hard. I think so too.

sindresorhus:
Moved to https://github.com/sindresorhus/conf/issues/86.

source/index.ts
Superjo149: @sindresorhus Looks correct to me as well. It will only exist if `options.schema` is present. The validate method already check its existence.

source/index.ts
papb: @sindresorhus Nowadays do you still recommend the `import foo = require('foo')` syntax for importing CJS-only modules?

Sorry for the clutter, but this was the "less worse" place to ask that I could think of. Thank you.

Issue #99, opened 2020-01-10
Feacture Request: ajv should have "removeAdditional: 'all' "  option


techo207:
when i set a schema to my conf file i want only accepted keys store in it. the new Ajv intance (line 84 in index.js ) should set "removeAdditional: 'all' " when a schema is given or conf should have a new option to only accept schema keys

sindresorhus:
This is better solved by https://github.com/sindresorhus/conf/issues/80.

PR #100, opened 2020-01-10
added restict key option to avoid undesirable keys

techo207:
according to https://github.com/sindresorhus/conf/issues/99 i made the modiffications to achive that


sindresorhus:
https://github.com/sindresorhus/conf/issues/99#issuecomment-586718038

PR #104, opened 2020-02-23
Move to TypeScript

### LINKED ISSUE #86: Improve the TypeScript types

Superjo149:
Got everything sorted out as requested. I did not add dot paths yet because we should get a good base first. That should be a seperate PR.

Everything is nicely typed now. The only thing that break for now, and I want your opinion on this first, is the schema. The schema now requires the enum JSONSchemaType for defining the types. I feel like that might be annoying for users, they will need to fix all their schemas. WDYT?

image

Fixes #86 

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#86: Improve the TypeScript types](https://issuehunt.io/repos/61670221/issues/86)
---

</details>


source/types.ts
```diff
@@ -0,0 +1,226 @@
...
+
+	__Currently this option doesn't work on Node.js 8 on macOS.__
+
+	@default false
+	*/
+	readonly watch?: boolean;
+}
+
+export type Migrations<T> = {
+	[key: string]: (store: Conf<T>) => void;
```
sindresorhus:
```suggestion
	[semverSpecifier: string]: (store: Conf<T>) => void;
```

source/conf.ts
```diff
@@ -0,0 +1,531 @@
...
+	/**
+	Watches the given `key`, calling `callback` on any changes.
+
+	@param key - The key wo watch.
+	@param callback - A callback function that is called on any changes. When a `key` is first set `oldValue` will be `undefined`, and when a key is deleted `newValue` will be `undefined`.
+	*/
+	onDidChange<TKey extends keyof T>(
+		key: TKey,
+		callback: OnDidChangeCallback<T[TKey]>
+	): () => void {
```
sindresorhus: What do you think about making the return function here a named type so it's clearer what it does?

source/conf.ts
```diff
@@ -0,0 +1,531 @@
...
+		};
+
+		this._events.on('change', onChange);
+		return () => this._events.removeListener('change', onChange);
+	}
+
+	private readonly _deserialize: Deserialize<T> = value => JSON.parse(value);
+	private readonly _serialize: Serialize<T> = value => JSON.stringify(value, null, '\t');
+
+	private _validate(data: T): void {
```
sindresorhus: Can you use private class fields instead for these? Now available in TS 3.8. Then we can also drop the `_` prefix.

source/conf.ts
```diff
@@ -0,0 +1,531 @@
...
+		}
+
+		if (semver.gt(candidateVersion, versionToMigrate)) {
+			return false;
+		}
+
+		return true;
+	}
+
+	private _get<TKey extends keyof T>(key: TKey, defaultValue?: T[TKey] | string): T[TKey] {
```
sindresorhus: Can you drop the `T` prefix from all the generic type names?

source/types.ts
```diff
@@ -0,0 +1,226 @@
...
+	projectName?: string;
+
+	/**
+	You only need to specify this if you don't have a package.json file in your project or if it doesn't have a version defined within it.
+
+	Default: The name field in the `package.json` closest to where `conf` is imported.
+	*/
+	projectVersion?: string;
+
+	/*
```
sindresorhus:
```suggestion
	/**
```

test/index.spec.ts
```diff
@@ -59,7 +65,7 @@ test('.set() - with undefined', t => {
 
 test('.set() - with unsupported values', t => {
 	t.throws(() => {
-		t.context.config.set('a', () => {});
+		t.context.config.set('a', () => { });
```
sindresorhus: Don't do unrelated changes.

package.json
```diff
@@ -39,27 +42,65 @@
...
+	"ava": {
+		"babel": false,
+		"compileEnhancements": false,
+		"extensions": [
+			"ts"
+		],
+		"require": [
+			"ts-node/register"
+		]
+	},
```
sindresorhus: Latest AVA version supports TypeScript natively. See: https://github.com/sindresorhus/got/commit/7f715719d92752c1f1838aa7d1c40d0537235767

package.json
```diff
@@ -39,27 +42,65 @@
...
+	"xo": {
+		"extends": "xo-typescript",
+		"extensions": [
+			"ts"
+		],
+		"rules": {
+			"@typescript-eslint/no-dynamic-delete": "off"
```
sindresorhus: Better to just disable it inline with a ESLint disable comment.

package.json
```diff
@@ -39,27 +42,65 @@
...
+		"extension": [
+			".ts"
+		],
+		"include": [
+			"source/**"
+		]
+	},
 	"dependencies": {
 		"ajv": "^6.10.2",
+		"crypto": "^1.0.1",
```
sindresorhus: This is a built-in Node.js core module and should not be here.

package.json
```diff
@@ -39,27 +42,65 @@
...
 		"delay": "^4.3.0",
+		"eslint-config-xo-typescript": "^0.22.0",
+		"nyc": "^14.1.1",
 		"p-event": "^4.1.0",
 		"tempy": "^0.3.0",
+		"ts-node": "^8.5.2",
 		"tsd": "^0.11.0",
-		"xo": "^0.24.0"
+		"typescript": "^3.7.5",
+		"xo": "^0.25.3"
```
sindresorhus: Can you bump all the dependencies to latest?

source/index.ts
```diff
@@ -0,0 +1,5 @@
+import {Conf} from './conf';
+
+export {Conf};
+
+export default Conf;
```
sindresorhus: I don't really see the point of this file? Why can't the `conf.ts` contents be in here?

source/types.ts
```diff
@@ -0,0 +1,226 @@
...
+	console.log(config.get('foo'));
+	//=> 50
+
+	config.set('foo', '1');
+	// [Error: Config schema violation: `foo` should be number]
+	```
+
+	**Note:** The `default` value will be overwritten by the `defaults` option if set.
+	*/
+	schema?: { [TProp in keyof T]: Schema; };
```
sindresorhus:
```suggestion
	schema?: {[Property in keyof T]: Schema};
```

source/types.ts
```diff
@@ -0,0 +1,226 @@
...
+				store.set('phase', '1.0.2');
+			},
+			'>=2.0.0': store => {
+				store.set('phase', '>=2.0.0');
+			}
+		}
+	});
+	```
+	*/
+	migrations?: Migrations<T>;
```
sindresorhus: All of these should be prefixed with `readonly`.

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+		"outDir": "dist",
+		"target": "es2017",
+		"lib": [
+			"es2017"
+		]
+	},
+	"include": [
+		"source/*.ts"
+	],
+	"exclude": ["node_modules", "dist"],
```
sindresorhus: This is moot. `node_modules` and the destination are excluded by default.

source/conf.ts
```diff
@@ -142,31 +157,260 @@ class Conf {
...
+		return this._handleChange(() => this.get(key), callback);
+	}
+
+	/**
+	Watches the whole config object, calling `callback` on any changes.
+
+	@param callback - A callback function that is called on any changes. When a `key` is first set `oldValue` will be `undefined`, and when a key is deleted `newValue` will be `undefined`.
+	*/
+	onDidAnyChange(
+		callback: OnDidChangeCallback<T>
```
sindresorhus: The callback here is not the same. See the existing index.d.ts file.

source/conf.ts
```diff
@@ -272,190 +516,16 @@ class Conf {
...
-
-	* [Symbol.iterator]() {
-		for (const [key, value] of Object.entries(this.store)) {
-			yield [key, value];
-		}
-	}
 }
 
-module.exports = Conf;
+export default Conf;
```
sindresorhus: Can you add this hack: https://github.com/sindresorhus/is/blob/05cdaccf924c37b54f4fdd149caab5ebe2af2f1b/source/index.ts#L655-L657 So CommonJS projects can import this without having to access Conf with `conf.default`.

source/conf.ts
```diff
@@ -142,31 +157,260 @@ class Conf {
...
 	}
 
-	_validate(data) {
+	/**
+	Get an item.
+
+	@param key - The key of the item to get.
+	@param defaultValue - The default value if the item does not exist.
+	*/
+	get<TKey extends keyof T>(key: TKey | string, defaultValue?: T[TKey] | any): T[TKey] | undefined {
```
sindresorhus: Don't use `any`. Use `unknown` or strongly-typed.

sindresorhus:
Make sure you look at and address all the feedback in the original PR. For example: https://github.com/sindresorhus/conf/pull/96#discussion_r362317644

sindresorhus:
> The schema now requires the enum JSONSchemaType for defining the types.

What changed that makes it require this?

Superjo149:
> > The schema now requires the enum JSONSchemaType for defining the types.
> 
> What changed that makes it require this?

Allright, nevermind. This is also a typing issue. Because it does support strings. I will look at your comments later today.

source/conf.ts
Superjo149: Yes, that would be cleaner 👍

source/types.ts
Superjo149: I'll fix this with `Readonly<>` otherwise I won't be able to set them internally in the constructor.

source/index.ts
Superjo149: This was some eslint thing, because we require Conf in `types.ts` as well. But it's a bit useless to have it in another file, so I'll place it back in index.ts

test/index.spec.ts
Superjo149: This is my code formatter :s

Superjo149:
All comments have been resolved. Tests seem to be fine, except for the `Failed to load plugin 'unicorn' declared in 'BaseConfig` error on Node 8.

sindresorhus:
I don't see https://github.com/sindresorhus/conf/pull/104#issuecomment-590239406 being handled.

sindresorhus:
All the methods should also use private class fields.

sindresorhus:
This is not done: https://github.com/sindresorhus/conf/pull/104#discussion_r383162007

sindresorhus:
https://github.com/sindresorhus/conf/pull/104#discussion_r383156125 is not resolved.

package.json
```diff
@@ -39,27 +43,63 @@
...
+				"test/": "dist/test/"
+			}
+		}
+	},
+	"xo": {
+		"extends": "xo-typescript",
+		"extensions": [
+			"ts"
+		]
+	},
```
sindresorhus: XO now supports TS built-in, so you can upgrade and drop this. See: https://github.com/xojs/xo/releases/tag/v0.27.0

tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+	"compilerOptions": {
+		"outDir": "dist",
+		"target": "es2018",
+		"lib": [
+			"es2018"
+		]
+	},
+	"include": [
+		"source",
+		"test"
```
sindresorhus: Because of https://github.com/sindresorhus/conf/pull/104/files#r383791352, `test` can be dropped here (don't forget to update `main` again in package.json).

source/types.ts
```diff
@@ -0,0 +1,233 @@
...
+export type ConfSchema<T> = { [Property in keyof T]: ValueSchema };
+export type ValueSchema = TypedJSONSchema;
+
+export type Serialize<T> = (value: T) => string;
+export type Deserialize<T> = (text: string) => T;
+
+export type OnDidChangeCallback<T> = (newValue?: T, oldValue?: T) => void;
+export type OnDidAnyChangeCallback<T> = (newValue?: Readonly<T>, oldValue?: Readonly<T>) => void;
+
+export type Unregister = () => EventEmitter;
```
sindresorhus:
```suggestion
export type Unsubscribe = () => EventEmitter;
```

source/types.ts
```diff
@@ -0,0 +1,233 @@
...
+	@default false
+	*/
+	readonly watch?: boolean;
+}
+
+export type Migrations<T> = {
+	[semverSpecifier: string]: (store: Conf<T>) => void;
+};
+
+export type ConfSchema<T> = { [Property in keyof T]: ValueSchema };
```
sindresorhus: This is the only type with `Conf` prefix, so I think we should drop it.
```suggestion
export type Schema<T> = {[Property in keyof T]: ValueSchema};
```

Superjo149:
@sindresorhus What should I do with the `node/no-unsupported-features` errors on `Rest/spread` for node 8 and `omission of catch binding` for node 10 rules? I think we still want to support these, right?

Superjo149:
Allright. I also made some of the anys better typed, except for the global `T` and `defaultValues`, because that will break more than it fixes, and its mostly internal use anyways.
I also went over everything again. Turns out I also forgot the overloading for `set` this is now also fixed and requires // @ts-ignore for the tests to compile which is good.

Superjo149:
Allright. I also made some of the anys better typed, I also managed to get `T` and `defaultValues`, typed, which was quit the hassle. I also went over everything again. Not using the anys at all there is very cumbersome, so there are still som lift in the T generic assignment.

Turns out I also forgot the overloading for `set` this is now also fixed and requires // @ts-ignore for the tests to compile which is good.
One thing with `set` and I don't think this was different with the old types. The nested properties do not work. It's either typesafety on the main object or allow nested properties and lose typesafety there.
image

What do you think we should do for this?

source/index.ts
```diff
@@ -374,7 +374,7 @@ export class Conf<T extends any = any> implements Iterable<[keyof T, T[keyof T]]
 
 	private _handleChange<Key extends keyof T>(
 		getter: () => T | T[Key] | undefined,
-		callback: OnDidAnyChangeCallback<T> | OnDidChangeCallback<T[Key]>
+		callback: OnDidAnyChangeCallback<any> | OnDidChangeCallback<any>
```
sindresorhus: Why `any`?

source/types.ts
```diff
@@ -225,7 +225,7 @@ export type Schema<T> = { [Property in keyof T]: ValueSchema };
 export type ValueSchema = TypedJSONSchema;
 
 export type Serialize<T> = (value: T) => string;
-export type Deserialize<T> = (text: string) => T;
+export type Deserialize<T> = (text: string) => T | { [key: string]: any };
```
sindresorhus: Why this change?

source/index.ts
```diff
@@ -180,8 +180,8 @@ export class Conf<T extends any = any> implements Iterable<[keyof T, T[keyof T]]
 	@param {key|object} - You can use [dot-notation](https://github.com/sindresorhus/dot-prop) in a key to access nested properties. Or a hashmap of items to set at once.
 	@param value - Must be JSON serializable. Trying to set the type `undefined`, `function`, or `symbol` will result in a `TypeError`.
     */
+	set<Key extends keyof T>(key: Key | string, value: T[Key] | unknown): void;
```
sindresorhus: This should be two different overloads. This now allows you to set using `Key`, but also the wrong value.

source/index.ts
```diff
@@ -43,13 +43,13 @@ const checkValueType = <Key = unknown>(key: string, value: Key): void => {
 const INTERNAL_KEY = '__internal__';
 const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
 
-export class Conf<T extends any = any> implements Iterable<[keyof T, T[keyof T]]> {
+export class Conf<T extends { [key: string]: any } = any> implements Iterable<[keyof T, T[keyof T]]> {
```
sindresorhus: Shouldn't it be:

```suggestion
export class Conf<T extends {[key: string]: any} = {[key: string]: any}> implements Iterable<[keyof T, T[keyof T]]> {
```

?

sindresorhus:
> The nested properties do not work. It's either typesafety on the main object or allow nested properties and lose typesafety there.

Yeah, I don't really know what to do about that. I guess all we can do is to clearly document in the readme/TS that it won't work when using TypeScript.

sindresorhus:
Bump

Superjo149:
I have had other priorities, I will pick this up in the week of the 20th of April

source/index.ts
Superjo149: It was because it was hard to type `.call`. But I found a solution.

source/types.ts
Superjo149: This is not need anymore.

source/index.ts
Superjo149: Yes, I initially had some issues with that. But it seems to be working now.

Superjo149:
I've dealt with the fixes. I have also applied all the changes that happend on master to the typescript version.

source/types.ts
```diff
@@ -0,0 +1,233 @@
...
+
+	/**
+	You only need to specify this if you don't have a package.json file in your project or if it doesn't have a version defined within it.
+
+	Default: The name field in the `package.json` closest to where `conf` is imported.
+	*/
+	projectVersion?: string;
+
+	/**
+	_Don't use this feature until [this issue](https://github.com/sindresorhus/conf/issues/92) has been fixed._
```
sindresorhus, line 76: This was removed. Can you go through each commit since this was opened to ensure that it's fully in sync with `master`?

package.json
```diff
@@ -39,27 +43,51 @@
...
 		"delay": "^4.3.0",
+		"eslint-config-xo-typescript": "^0.26.0",
+		"nyc": "^15.0.0",
 		"p-event": "^4.1.0",
-		"tempy": "^0.3.0",
+		"tempy": "^0.4.0",
+		"ts-node": "^8.6.2",
 		"tsd": "^0.11.0",
-		"xo": "^0.24.0"
+		"typescript": "^3.8.3",
```
sindresorhus, line 90:
```suggestion
		"typescript": "3.9.3",
```

source/index.ts
```diff
@@ -0,0 +1,576 @@
...
+
+	if (nonJsonTypes.includes(type)) {
+		throw new TypeError(`Setting a value of type \`${type}\` for key \`${key}\` is not allowed as it's not supported by JSON`);
+	}
+};
+
+const INTERNAL_KEY = '__internal__';
+const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
+
+export class Conf<T extends { [key: string]: any } = { [key: string]: any }> implements Iterable<[keyof T, T[keyof T]]> {
```
sindresorhus, line 46:
```suggestion
export class Conf<T extends Record<string, any> = Record<string, unknown>> implements Iterable<[keyof T, T[keyof T]]> {
```

source/types.ts
Superjo149: I did already go through everything on master. I just did not check the `index.d.ts` since it is auto generated on my end. I'll recheck to make sure, but it should be good besides that.

sindresorhus:
Tests are not passing. Try removing `xo` from https://github.com/sindresorhus/conf/pull/104/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R18 and run `npm test` locally and ensure it passes.

source/index.ts
```diff
@@ -0,0 +1,576 @@
...
+
+	/**
+    Get an item.
+
+	@param key - The key of the item to get.
+	@param defaultValue - The default value if the item does not exist.
+	*/
+	get<Key extends keyof T>(key: Key | string): T[Key] | undefined;
+	get<Key extends keyof T, Default = unknown>(key: Key | string, defaultValue?: Default): T[Key] | Default;
+	get<Key extends keyof T, Default = unknown>(key: Key | string, defaultValue?: Default): T[Key] | Default | undefined {
```
sindresorhus, line 169: `Default` needs to be constrained to be the same type as `Key`. Then you can also just return `Default`.

source/index.ts
```diff
@@ -0,0 +1,576 @@
...
+			if (!options.projectVersion) {
+				throw new Error('Project version could not be inferred. Please specify the `projectVersion` option.');
+			}
+
+			this._migrate(options.migrations, options.projectVersion);
+		}
+	}
+
+	/**
+    Get an item.
```
sindresorhus, line 162: Use tab-indentaiton.

Superjo149:
I think it's ok now. The tests weren't working, implemented them like in https://github.com/sindresorhus/got/. The linter issues are fixed and I improved the typing for .get

sindresorhus:
Can you add a type test to confirm https://github.com/sindresorhus/conf/pull/104#discussion_r429622319 is working?

sindresorhus:
Travis is still failing.

Superjo149:
Allright, looked at it again and fixed the problem. But there are still some issues.
1. The `xo` package requires `get-stdin` which has a `await`, so either we have to upgrade the node version or downgrade `xo`
2. Then, I have this on my mac as well. The 2 watch tests fail. I debugged this to being `fs.watch` not working properly. This does work when using `fs.watchFile`.

sindresorhus:
1. I plan to require Node.js 10 after this PR.
2. This is failing in master too. You don't have to worry about that. Probably related to https://github.com/nodejs/node/issues/29460

sindresorhus:
Thanks for working on this! Looks great now 🙌🏻

Superjo149:
Ah allright, glad I could help. I will be able to benefit from this as well in my app 🙂

package.json
```diff
@@ -11,14 +11,17 @@
...
 	"engines": {
-		"node": ">=8"
+		"node": ">=8.3.0"
 	},
 	"scripts": {
-		"test": "xo && ava && tsd"
+		"test": "xo && tsc --noEmit && nyc ava",
+		"prepare": "npm run build",
+		"build": "del dist && tsc",
+		"prepublishOnly": "npm run build"
```
fregante, line 20: As far as I know `prepare` is meant to replace `prepublishOnly` so you don't need both.

However I've been using the more specific `prepack` which feels like it's the exact moment when files should be built: right before the tar.gz is created.

sindresorhus: > As far as I know prepare is meant to replace prepublishOnly so you don't need both.

Yes, but it's a mess since old npm version use `prepare`, new ones use `prepublishOnly`, and future ones will use `prepare` but with it will work differently. I'm not sure why we use both here. Maybe look at the git commits/blame.

> However I've been using the more specific prepack which feels like it's the exact moment when files should be built: right before the tar.gz is created.

Does that one also work when users install a Git commit hash directly? `npm i sindresorhus/conf#ad32d2`

fregante: > prepack: run BEFORE a tarball is packed (on npm pack, npm publish, and when installing git dependencies)

fregante: Compared with:

> prepare: Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below). This is run AFTER prepublish, but BEFORE prepublishOnly.

sindresorhus: https://github.com/sindresorhus/conf/commit/be694026e6fa09a3482a4d6fd3f260f07d1e36f6

PR #117, opened 2020-07-26
Fix the `.get()` TypeScript return type

### LINKED ISSUE #116: TypeScript types should take into account defaults
https://github.com/sindresorhus/electron-store/blob/966a9c736318c6c51083de08e5eb5d1276e00047/index.test-d.ts#L47 should not be `number | undefined` as it already has a default value defined in the constructor: https://github.com/sindresorhus/electron-store/blob/966a9c736318c6c51083de08e5eb5d1276e00047/index.test-d.ts#L40-L43

SnosMe:
Closes #116

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -167,8 +167,8 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	*/
 	get<Key extends keyof T>(key: Key): T[Key];
 	get<Key extends keyof T>(key: Key, defaultValue: Required<T>[Key]): Required<T>[Key];
-	// This overload is used to access via dot-notation
-	// Exclude `keyof T`, incorect type for default value should not fallthrough to this overload
+	// This overload is used for dot-notation access.
+	// We exclude `keyof T` as an incorrect type for the default value should not fall through to this overload.
 	get<Key extends string, Value = unknown>(key: Exclude<Key, keyof T>, defaultValue?: Value): Value;
 	get(key: string, defaultValue?: unknown): unknown {
 		if (this.#options.accessPropertiesByDotNotation) {
--- a/source/index.ts
+++ b/source/index.ts
@@ -166,9 +166,8 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	@param defaultValue - The default value if the item does not exist.
 	*/
 	get<Key extends keyof T>(key: Key): T[Key];
-	get<Key extends string, TValue = unknown>(key: Key extends keyof T ? never : Key): TValue;
 	get<Key extends keyof T>(key: Key, defaultValue: Required<T>[Key]): Required<T>[Key];
-	get<Key extends string, TValue = unknown>(key: Key extends keyof T ? never : Key, defaultValue: TValue): TValue;
+	get<Key extends string, TValue = unknown>(key: Key extends keyof T ? never : Key, defaultValue?: TValue): TValue;
 	get(key: string, defaultValue?: unknown): unknown {
 		if (this.#options.accessPropertiesByDotNotation) {
 			return this._get(key, defaultValue);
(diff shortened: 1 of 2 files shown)


source/index.ts
```diff
@@ -165,9 +165,11 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	@param key - The key of the item to get.
 	@param defaultValue - The default value if the item does not exist.
 	*/
-	get<Key extends keyof T>(key: Key | string): T[Key] | undefined;
-	get<Key extends keyof T, Default = T[Key]>(key: Key | string, defaultValue: Default): T[Key] | Default;
-	get<Key extends keyof T, Default = T[Key]>(key: Key | string, defaultValue?: Default): Default | undefined {
+	get<Key extends keyof T>(key: Key): T[Key];
+	get<T = unknown>(key: string): T;
+	get<Key extends keyof T, Default = Required<T>[Key]>(key: Key, defaultValue: Default): Required<T>[Key] | Default;
```
sindresorhus, line 170: The return value and the default value should always be the same type.

source/index.ts
```diff
@@ -165,9 +165,11 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
...
+	get<T = unknown, Default = T>(key: string, defaultValue: Default): T | Default;
+	get(key: string, defaultValue?: any): any {
```
sindresorhus, line 172: Don't use `any`

test/index.test-d.ts
```diff
@@ -127,6 +127,8 @@ expectType<boolean>(config.get('isRainbow', false));
 
 expectType<string | undefined>(config.get('unicorn'));
 expectType<string>(config.get('unicorn', 'rainbow'));
+// @ts-expect-error
+expectType<number>(config.get('unicorn', 1));
```
sindresorhus, line 131: `tsd` has a `expectError` assertion.

source/index.ts
```diff
@@ -165,9 +165,10 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	@param key - The key of the item to get.
 	@param defaultValue - The default value if the item does not exist.
 	*/
-	get<Key extends keyof T>(key: Key | string): T[Key] | undefined;
-	get<Key extends keyof T, Default = T[Key]>(key: Key | string, defaultValue: Default): T[Key] | Default;
-	get<Key extends keyof T, Default = T[Key]>(key: Key | string, defaultValue?: Default): Default | undefined {
+	get<Key extends keyof T>(key: Key): T[Key];
+	get<Key extends keyof T>(key: Key, defaultValue: Required<T>[Key]): Required<T>[Key];
+	get<Key extends string, TValue = unknown>(key: Key extends keyof T ? never : Key, defaultValue?: TValue): TValue;
```
sindresorhus, line 170:
```suggestion
	get<Key extends string, Value = unknown>(key: Key extends keyof T ? never : Key, defaultValue?: Value): Value;
```

sindresorhus: Could you add a short code comment about the purpose of this overload? It's not immediately clear.

SnosMe: Its purpose is commit name: using incorect type for default value should not fallthrough to overload with unknown type.

If you want to put it in source I'm not sure how to do it elegantly 😄

sindresorhus: Just write what you wrote above, just in a code comment.

test/index.test-d.ts
```diff
@@ -67,7 +67,7 @@ conf.set({
 	}
 });
 
-expectType<string | undefined>(conf.get('foo'));
+expectType<string>(conf.get('foo'));
```
Superjo149, line 70: The second one was indeed incorrect. But Is it correct to assume that a value will never be undefined when we do not pass a default? I think we should still make it possible to be undefined to encourage people to handle an undefined state.

SnosMe: Yes, value will never be undefined unless developer make a bug such in this test, type `UnicornFoo` is source of truth.
```ts
type UnicornFoo = {
	foo: string;
};
const conf = new Conf<UnicornFoo>({accessPropertiesByDotNotation: true});
```
Maybe we should validate `default` option, but it is out of scope of this PR.
Like this
```ts
	defaults?: Readonly<{
		[P in keyof T]: Exclude<T[P], undefined>
	}>;
```

sindresorhus: > Maybe we should validate default option, but it is out of scope of this PR.

That would be nice, yes, but I agree, it should be a separate PR.

sindresorhus:
Thanks for fixing this, @SnosMe 🙌🏻

PR #119, opened 2020-07-28
Use typescript `private` instead of `#private`

### LINKED ISSUE #118: Use of #private forces Typescript users to target es2015 or later, use 3.8 or later
Found on Definitely Typed, where kap-plugin depends on conf. By default, Definitely Typed tsconfigs use the default target, which is (still) es5. This fails since ES5 doesn't have WeakMap, which `#private` emit depends on.

I would not personally use `#private` in conf:

1.  It doesn't have any information it needs to hide from other code, it just needs to prevent mistaken accesses to its internal details. 
2. The emitted Javascript is super weird, using WeakMaps unless you target ESNext.
2. If you do target ESNext, Firefox and Safari don't support `#private`.
3. Only Typescript 3.8 and above support `#private` at all. 

I think Typescript-classic `private` is the right choice.

sandersn:
Fixes #118

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -46,10 +46,10 @@ const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
 class Conf<T extends Record<string, any> = Record<string, unknown>> implements Iterable<[keyof T, T[keyof T]]> {
 	readonly path: string;
 	readonly events: EventEmitter;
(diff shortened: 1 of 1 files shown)


sindresorhus:
https://github.com/sindresorhus/conf/issues/118#issuecomment-665229257

PR #121, opened 2020-08-10
Adding schema as a variable will not work

kevinsimper:
It will complain if you add it like this, but not if you define it inside.

I have no idea why, I don't know enough about typescript to know the difference.

### DIFF
--- a/test/index.test-d.ts
+++ b/test/index.test-d.ts
@@ -55,6 +55,30 @@ new Conf<UnicornFoo>({
 	}
 });
 
+const schema = {
+	foo: {
+		type: 'string',
+		default: 'foobar'
+	},
+	unicorn: {
+		type: 'boolean'
+	},
+	hello: {
+		type: 'number'
+	},
+	nested: {
+		type: 'object',
+		properties: {
+			prop: {
+				type: 'number'
+			}
+		}
+	}
+}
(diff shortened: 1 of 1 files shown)


sindresorhus:
As far as I know, this is how TS works, you have to import the `Schema` type and explicitly type the object literal.

sindresorhus:
// @Superjo149

kevinsimper:
@sindresorhus The weird thing is that the example in the readme does not work with how the types is, but I can't say why. I only copied the example index.test-d.ts and split it up with the variable schema.

Superjo149:
Fixed in #122

SnosMe:
@kevinsimper You should declare schema object as const

```ts
const schema = {
	...
} as const
  ^^^^^^^^ available since TS 3.4
```

#122 isn't needed.

Superjo149:
> @kevinsimper You should declare schema object as const
> 
> ```ts
> const schema = {
> 	...
> } as const
>   ^^^^^^^^ available since TS 3.4
> ```
> 
> #122 isn't needed.

Aha, nice one.

kevinsimper:
@SnosMe Thank you, did not know that 👍 But I don't understand why it would not be const when defined as a const

Issue #109, opened 2020-04-29
How to handle arrays?


bloodchen:
Let's say I have a config

```js
{
   "items":[ {"name":"abc"},{"name":"dadd"}]
}
```

How can I add an item to items array?

Thanks

sindresorhus:
There's currently no nice way to do this. You'll have to do something like:

```js
store.set('items', [...store.get('items'), newItem]);
```

I plan to add a convenience method for this. See: https://github.com/sindresorhus/electron-store/issues/52

Nisthar:
@sindresorhus Doesn't this cause performance issues when the array is big?

sindresorhus:
Only for huge arrays, and you should not store huge chunks of data in `conf` anyway.

sindresorhus:
For anyone wanting to work on this, see the initial attempt and feedback in: https://github.com/sindresorhus/conf/pull/124

PR #124, opened 2020-08-20
Add `.appendToArray()` method

subins2000:
* Addresses https://github.com/sindresorhus/electron-store/issues/52
* Addresses #109

Allows appending or creating an array item with `.appendToArray` :

```
store.appendToArray('foo', 'a') // foo will be ['a']
store.appendToArray('foo', 'b') // foo will be ['a', 'b']
```

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -166,7 +166,7 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	@param defaultValue - The default value if the item does not exist.
 	*/
 	get<Key extends keyof T>(key: Key): T[Key];
-	get<Key extends keyof T>(key: Key, defaultValue: Required<T>[Key]): Required<T>[Key];
+	get<Key extends keyof T>(key: Key, defaultValue: Required<T>[Key] | Required<unknown>): Required<T>[Key];
 	// This overload is used for dot-notation access.
 	// We exclude `keyof T` as an incorrect type for the default value should not fall through to this overload.
 	get<Key extends string, Value = unknown>(key: Exclude<Key, keyof T>, defaultValue?: Value): Value;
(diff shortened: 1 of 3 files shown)


test/index.ts
```diff
@@ -99,6 +99,17 @@ test('.set() - invalid key', t => {
 	}, {message: 'Expected `key` to be of type `string` or `object`, got number'});
 });
 
+test('.appendToArray()', t => {
```
sindresorhus, line 102: This needs an assertion for the case where the value is already set to a different type.

I would also appreciate more assertions if you can try to think of some more.

source/index.ts
```diff
@@ -223,6 +223,25 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 		this.store = store;
 	}
 
+	/**
```
sindresorhus, line 226: The types here should limit it to only allow using `appendToArray` on a key defined to have an array value or unknown.

source/index.ts
```diff
@@ -223,6 +223,25 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
...
+    Append an item to array
+
+	@param {key|object} - You can use [dot-notation](https://github.com/sindresorhus/dot-prop) in a key to access nested properties. Or a hashmap of items to set at once.
+	@param value - Must be JSON serializable. Trying to set the type `undefined`, `function`, or `symbol` will result in a `TypeError`.
+    */
```
sindresorhus, line 231:
```suggestion
*/
```

source/index.ts
```diff
@@ -223,6 +223,25 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
...
+	appendToArray<Key extends keyof T>(key: Key, value: Required<T>[Key]): void;
+	// This overload is used for dot-notation access.
+	appendToArray<Key extends string>(key: Key, value: Required<T>[Key]): void {
+		const list = this.get(key, [] as T[Key]);
+
+		if (!Array.isArray(list)) {
+			throw new TypeError('Config param already set and is not array');
```
sindresorhus, line 238:
```suggestion
			throw new TypeError(`The key \`${key}\` is already set to a non-array value`);
```

source/index.ts
```diff
@@ -223,6 +223,25 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
...
+		}
+
+		list.push(value);
+		this.set(key, list);
```
sindresorhus, line 242:
```suggestion
		this.set(key, [...list, value]);
```

source/index.ts
```diff
@@ -223,6 +223,25 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
...
+	/**
+    Append an item to array
+
+	@param {key|object} - You can use [dot-notation](https://github.com/sindresorhus/dot-prop) in a key to access nested properties. Or a hashmap of items to set at once.
+	@param value - Must be JSON serializable. Trying to set the type `undefined`, `function`, or `symbol` will result in a `TypeError`.
+    */
+	appendToArray<Key extends keyof T>(key: Key, value: Required<T>[Key]): void;
+	// This overload is used for dot-notation access.
+	appendToArray<Key extends string>(key: Key, value: Required<T>[Key]): void {
+		const list = this.get(key, [] as T[Key]);
```
sindresorhus, line 235:
```suggestion
		const array = this.get(key, [] as T[Key]);
```

sindresorhus (changes requested):

sindresorhus:
It must be documented in the readme.

source/index.ts
subins2000: I tried many, but can't figure it out. The best I could come up with was
```ts
appendToArray<Key extends keyof T, Index extends keyof any[]>(key: Key, value: T[Key][Index]): void;
```

subins2000: @sindresorhus I'm not able to make it, would like to know the solution please

LanceGundersen:
This is exactly what I am looking for! Anything I can help to get this merged?

sindresorhus:
> Anything I can help to get this merged?

Yes, see: https://github.com/sindresorhus/conf/pull/124#discussion_r480275958

sindresorhus:
@subins2000 Bump

PR #136, opened 2021-01-07
Fix `.clear()` behavior

### LINKED ISSUE #135: reset without arguments (keys)
Currently, calling `MyStore.reset()` without any arguments is a noop (without any warnings), contrary to what I would expect (and I imagine others). At the very least, perhaps a basic solution to just that problem:
1. Warn if no key are provided
2. Type should require at least one key, e..g `reset(key: string, ...keys: string[])`

Furthermore, my expectation was that calling reset without arguments would reset all the keys. 
E.g.
```ts
const keys = Object.keys(MyStore) as (keyof (typeof MyStore.store))[]
MyStore.reset(...keys)
```
However even this isn't correct, because it would only reset those keys in the current store value, not all keys (e.g. if there were some keys undefined, they would not be reset to their defaults).

**What I am most interested in is some method `resetAll()` that effectively resets the entire store to be equal to the originally provided value to defaults, with no additional properties.**

----
```ts
const defaults = { a: 1, b: 2 }
MyStore = new Conf({ ..., defaults }) // current value is { a: 1, b: 2 }
MyStore.store = { c: 3 } // current value is { a: 3 }
MyStore.reset() // noop <-- unexpected
MyStore.reset(...Object.keys(MyStore.store)) // current value is { }  <-- unexpected
MyStore.reset(...Object.keys(defaults)) //  current value is { c: 3} <-- unexpected
MyStore.store = defaults //  current value is { a: 1, b: 2 } which is what I would expect MyStore.reset() to do
```

sindresorhus:
Fixes #135

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -273,11 +273,9 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	This resets known items to their default values, if defined by the `defaults` or `schema` option.
 	*/
 	clear(): void {
-		const keys = Object.keys(this.store);
-
 		this.store = createPlainObject();
 
-		for (const key of keys) {
+		for (const key of Object.keys(this.#defaultValues)) {
 			this.reset(key);
 		}
 	}
--- a/source/index.ts
+++ b/source/index.ts
@@ -238,6 +238,8 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	/**
 	Reset items to their default values, as defined by the `defaults` or `schema` option.
 
+	@see `clear()` to reset all items.
+
 	@param keys - The keys of the items to reset.
 	*/
 	reset<Key extends keyof T>(...keys: Key[]): void {
@@ -267,9 +269,17 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 
 	/**
 	Delete all items.
+
+	This resets known items to their default values, if defined by the `defaults` or `schema` option.
 	*/
 	clear(): void {
+		const keys = Object.keys(this.store);
+
 		this.store = createPlainObject();
+
+		for (const key of keys) {
+			this.reset(key);
+		}
 	}
 
 	/**
--- a/readme.md
+++ b/readme.md
@@ -292,6 +292,8 @@ Get an item or `defaultValue` if the item does not exist.
 
 Reset items to their default values, as defined by the `defaults` or `schema` option.
 
+Use `.clear()` to reset all items.
+
 #### .has(key)
 
 Check if an item exists.
@@ -303,6 +305,8 @@ Delete an item.
 #### .clear()
 
 Delete all items.
+
+This resets known items to their default values, if defined by the `defaults` or `schema` option.
 
 #### .onDidChange(key, callback)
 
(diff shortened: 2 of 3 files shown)


source/index.ts
```diff
@@ -267,9 +269,17 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 
 	/**
 	Delete all items.
+
+	This resets known items to their default values, if defined by the `defaults` or `schema` option.
 	*/
 	clear(): void {
+		const keys = Object.keys(this.store);
```
willium, line 276: If the current value of the store has a key deleted that is in the defaults or schema, I believe it wouldn't be included in this and you'd end up with an incomplete default.

source/index.ts
```diff
@@ -267,9 +269,17 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
...
+
 		this.store = createPlainObject();
+
+		for (const key of keys) {
+			this.reset(key);
+		}
```
willium, line 282: I think it'd make sense to emit the change event here, what do you think?

this.events.emit('change');

willium (commented):
Awesome to see!

source/index.ts
sindresorhus: That is already done, since we set `this.store`: https://github.com/sindresorhus/conf/blob/43c644f6fb0cd96a9c2e1f49b2ee0780cd089b4c/source/index.ts#L344

source/index.ts
sindresorhus: Good catch: https://github.com/sindresorhus/conf/pull/136/commits/8b1ee0b6eb2f5ff4226baa23dac3b1e44b77cf52

willium:
excited to see this land! only remaining concern is that this is probably a breaking change for people who rely on `clear()` to wipe the slate, ignoring defaults. It's your library, so up to you on how to handle that. If you want to keep it graceful, you could change the method signature to something like
```ts
public clear(restoreDefaults?: boolean): void
```
so that people like me can just pass `true` and current users won't notice any change

sindresorhus:
It's really just a bug fix, but I will do a major release just to be careful.

source/index.ts
willium: @sindresorhus this is actually causing a regressive bug in my program. While we set this.store, we do it before reseting the parameters, and we set it to a blank object—which disobeys the type contract. Subsequently it's emitting an onDidAnyChange event where the current value is undefined (which disobeys the type guarantees for the store, given defaults)

willium: this doesn't cause a static typing violation because of the use Partial in several places where I don't believe it makes sense. Ideally it'd be possible to create a Conf with strict type guarantee. The type of details should statically guaranteed to be the base type.

sindresorhus: I don't have the time to look into this right now, so it would be good if you could open an issue instead. Even better if you could also submit some failing tests.

willium: I created an [issue](https://github.com/sindresorhus/conf/issues/138). The type safety bug here is pretty important. The clear() bug is as well, given that is breaks the contract with the client regarding types.

Issue #114, opened 2020-07-05
[BUG] Assigning `conf.store` overrides applied migrations


ixrock:
package.json
`
"conf": "^7.0.1",
`

_Example:_

```js
import Config from "conf"

const config = new Config({
 migrations: {
    "2.0.0": store => store.set("lastSeenVersion", "2.0.0")
  }
})
config.store = {newData: "blabla"} // this will destroy marked migration in __internal__
```

sindresorhus:
// @rafaelramalho19 In case you have time/interest in looking into this. If not, no worries.

willium:
Was this resolved? Seems tricky.

sindresorhus:
If anyone wants to work on this, see the initial attempt and feedback in: https://github.com/sindresorhus/conf/pull/137

PR #137, opened 2021-01-11
Provide way to omit __interal__ from store on get

willium:
The internal key causes some problems, both during write (https://github.com/sindresorhus/conf/issues/114) and at read (can't trust MyStore.keys() or MyStore.values() without omitting __internal__). We should at least export the variable so the client can manually omit it (and at best, provide a more ergonomic pattern—perhaps a WeakMap?)

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -40,8 +40,8 @@ const checkValueType = (key: string, value: unknown): void => {
 	}
 };
 
-const INTERNAL_KEY = '__internal__';
-const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
+export const INTERNAL_KEY = '__internal__';
+export const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
 
 class Conf<T extends Record<string, any> = Record<string, unknown>> implements Iterable<[keyof T, T[keyof T]]> {
 	readonly path: string;


willium:
In my subclass, I do this:
```
	public get current(): T {
		// per this issue: https://github.com/sindresorhus/conf/pull/137
		// __internal__ causes some issues
		return omit(this.store, "__internal__") as T
	}

	public set current(value: T) {
		const { __internal__ } = this.store
		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
		this.store = {...value, __internal__ } as T
	}
```
which seems to work but isn't ideal (and at the very least, I'd love import INTERNAL_KEY rather than hardcode it here).

sindresorhus:
I'm not interested in exposing internal implementation details. This should be properly fixed by making changes to `this.store` not change `__internal__`. This can be done by making `this.store` be a facade for an internal `this._store` and have `this.store` be getter/setters that handles ignoring `__internal__`.

sindresorhus:
> (and at best, provide a more ergonomic pattern—perhaps a WeakMap?)

Cannot use `WeakMap` as the `__internal__` key needs to be persisted.

willium:
that would great. do you think the changes to the getter/setter goes beyond what I did in my subclass?

sindresorhus:
I think that should be enough as `this.store` is used for any reading/writing.

willium:
should I submit a PR with this change?

sindresorhus:
Yes, that would be great. Needs some tests too.

willium:
It's unfortunately not going to be super simple as migrations (or at least their tests) rely on being able to lookup migrations by the __internal__ property (for example [this](https://github.com/sindresorhus/conf/blob/master/test/index.ts#L865), relies on .get() which in turn uses .store). I'm not sure your preferred strategy for exposing/mocking private methods for tests, but if you're willing to remove the ability to .get() data off the internal key, we could introduce some kind of method _getInternal() that doesn't use .get store

willium:
We could also do some [`any` hacking](https://stackoverflow.com/a/35991491/347727) in the tests

sindresorhus:
> we could introduce some kind of method _getInternal() that doesn't use .get store, though in order to use it in the tests it'd have to be public...

Sounds good. We can make it public only when `if (process.env.NODE_ENV === 'test') {`.

sindresorhus:
Friendly bump :)

sindresorhus:
Closing for lack of activity.

PR #143, opened 2021-02-08
Fix issue with `onDidChange` not always being called

### LINKED ISSUE #108: onDidChange is called once
Replace fs.watch to fs.watchFile in _watch method.

nhevia:
Fixes #108 and fixes https://github.com/sindresorhus/electron-store/issues/93

Adds a configuration option called `watchFile`.

Works exactly like `watch` but uses `fs.watchFile` instead of `fs.watch`, allowing the subscription to read changes in the file more than once.

```jsx
useEffect(()=>{
    store.onDidChange('myKey', (newV, oldV) => {
      console.log(`Changed from ${oldV} to ${newV}`)
    })
},[])
```

image

Clearly slower (sometimes, benchmarking is unreliable 100/2000ms) than `watch` but the only way to truly watch a file in React (and possible other scenarios, but not tested).

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -848,9 +848,7 @@ test('.delete() - without dot notation', t => {
 	t.deepEqual(configWithoutDotNotation.get('foo.bar.zoo'), {awesome: 'redpanda'});
 });
 
-const test2 = process.platform === 'darwin' ? test.failing : test;
-
-test2('`watch` option watches for config file changes by another process', async t => {
+test('`watch` option watches for config file changes by another process', async t => {
 	const cwd = tempy.directory();
 	const conf1 = new Conf({cwd, watch: true});
 	const conf2 = new Conf({cwd});
@@ -877,7 +875,7 @@ test2('`watch` option watches for config file changes by another process', async
 	await pEvent(_events, 'change');
 });
 
-test2('`watch` option watches for config file changes by file write', async t => {
+test('`watch` option watches for config file changes by file write', async t => {
 	const cwd = tempy.directory();
 	const conf = new Conf({cwd, watch: true});
 	conf.set('foo', '🐴');
@@ -892,7 +890,7 @@ test2('`watch` option watches for config file changes by file write', async t =>
 	conf.onDidChange('foo', checkFoo);
 
 	(async () => {
-		await delay(5000);
+		await delay(7000);
 		fs.writeFileSync(path.join(cwd, 'config.json'), JSON.stringify({foo: '🦄'}));
 	})();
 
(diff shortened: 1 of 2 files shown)


ali-farooq0:
Great work @nhevia! Really need this to fix https://github.com/sindresorhus/electron-store/issues/93. @sindresorhus can we please get this in ASAP?

sindresorhus:
So `fs.watch` only fails on Ubuntu, right? Can't we just use `fs.watchFile` on Linux? No option needed.

sindresorhus:
Also please open an issue on Node.js. I want to ensure the issue will be eventually fixed in Node.js.

ali-farooq0:
> So `fs.watch` only fails on Ubuntu, right? Can't we just use `fs.watchFile` on Linux? No option needed.

I'm seeing https://github.com/sindresorhus/conf/issues/108 on macOS.

sindresorhus:
> I'm seeing #108 on macOS.

Ok, so let's use `fs.watch` on macOS too then.

ali-farooq0:
you mean `fs.watchFile`? Sure, works for me 👍 . @nhevia are you able to make the necessary changes? It will be a lot nicer to have this work without having to specify an extra option :).

sindresorhus:
Yes

nhevia:
Changes added.

Tests were failing, added to review.

ali-farooq0 (approved):
Wicked 👍

ali-farooq0:
Thanks for the quick turn around on this @nhevia!

test/index.ts
```diff
@@ -848,7 +848,8 @@ test('.delete() - without dot notation', t => {
 	t.deepEqual(configWithoutDotNotation.get('foo.bar.zoo'), {awesome: 'redpanda'});
 });
 
-const test2 = process.platform === 'darwin' ? test.failing : test;
+/* eslint no-negated-condition: 0 */
```
sindresorhus, line 851: The lint rule is there for a reason. Don't disable it.

test/index.ts
```diff
@@ -848,7 +848,8 @@ test('.delete() - without dot notation', t => {
...
+const test2 = process.platform !== 'win32' ? test.failing : test;
```
sindresorhus, line 852: Shouldn't the changes in this PR cause the test to pass on all OSes?

sindresorhus:
The PR title is outdated.

ali-farooq0:
Will you have time to address the feedback @nhevia?

sindresorhus:
@nhevia Bump

nhevia:
Tests passing. Feel free to squash this mess 👀

PR #147, opened 2021-03-25
Use module.children for parentDir, instead of module.parent

anujdatar:
module.parent is deprecated in node 14.6+


sindresorhus:
I plan to remove this code soon as it won't work with ESM. See: https://github.com/sindresorhus/conf/issues/150

PR #154, opened 2021-09-27
Support non-primitive values

Richienb:
Continuation of https://github.com/sindresorhus/electron-store/pull/171


sindresorhus:
I'm unfortunately going to pass on this. This change is too V8-specific and could become a maintenance nightmare if V8 changes the format or if we need to supprot other platforms in the future. Basically not worth it. Users can serialize themselves.

Issue #157, opened 2021-11-18
Add a `mode` option


ncallaway:
Add a `mode` option that will be used for writing the config file. This feature may be useful if you want to produce a config file that is not world or group readable (or writable).

An example use case is the `aws cli`, which allows the user to configure a profile with `aws configure --profile [profilename]`. The profiles are configured in `~/.aws/config` and `~/.aws/credentials` with a `0o600` mode.

sindresorhus:
Fixed by https://github.com/sindresorhus/conf/pull/158

PR #158, opened 2021-11-18
Add `configFileMode` option

ncallaway:
This is a proposed implementation for #157. This adds a `mode` option, which is passed through to the underlying call to `fs.writeFileSync` or `atomically.writeFileSync`.

The default mode here (`0o666`) is the default mode for both `fs.writeFileSync` and `atomically.writeFileSync` so this should be a backwards-compatible change.

### DIFF
--- a/source/types.ts
+++ b/source/types.ts
@@ -138,11 +138,15 @@ export interface Options<T> {
 	clearInvalidConfig?: boolean;
 
 	/**
-	The mode that will be used for the config file.
+	The [mode](https://en.wikipedia.org/wiki/File-system_permissions#Numeric_notation) that will be used for the config file.
+
+	You would usually not need this, but it could be useful if you want to restrict the permissions of the configuration file. Setting a permission such as 0o600 would result in a configuration file that can only be accessed by the user running the program.
+
+	Note that setting restrictive permissions can cause problems if multiple different users need to read the file.
 
 	@default 0o666
 	*/
-	mode: number | string;
+	configFileMode: number;
 
 	/**
 	Function to serialize the config object to a UTF-8 string when writing the config file.
--- a/source/types.ts
+++ b/source/types.ts
@@ -212,7 +212,7 @@ export interface Options<T> {
 	@default false
 	*/
 	readonly watch?: boolean;
-	
+
 	/**
 	The [mode](https://en.wikipedia.org/wiki/File-system_permissions#Numeric_notation) that will be used for the config file.
 
--- a/source/types.ts
+++ b/source/types.ts
@@ -138,6 +138,13 @@ export interface Options<T> {
 	clearInvalidConfig?: boolean;
 
 	/**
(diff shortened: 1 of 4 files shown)


source/types.ts
```diff
@@ -137,6 +137,13 @@ export interface Options<T> {
 	*/
 	clearInvalidConfig?: boolean;
 
+	/**
+	The mode that will be used for the config file.
+
+	@default 0o666
+	*/
+	mode: number | string;
```
sindresorhus, line 145:
```suggestion
	configFileMode: number;
```

I'm aware the underlying API supports a string, but that's only for legacy reasons. I prefer to keep it simple.

source/types.ts
```diff
@@ -137,6 +137,13 @@ export interface Options<T> {
 	*/
 	clearInvalidConfig?: boolean;
 
+	/**
+	The mode that will be used for the config file.
```
sindresorhus, line 141: Linkify `mode` to a website that explains what `mode` is. Not every user is familiar with the concept.

source/types.ts
```diff
@@ -137,6 +137,13 @@ export interface Options<T> {
...
+
```
sindresorhus, line 142: The docs should explain when it would be useful. Most devs would not need this, and from experience, setting `mode` can cause a lot of problems. For example, user run a tool with `sudo` and then later not with `sudo` and then the config file is not readable, etc. I did the mode thing in a different config library, and it caused endless of support requests.

For example, on macOS, `conf` by default, stores the config in `~/Library/Preferences`, which is not world writable. So this would not be needed if you target only macOS.

https://github.com/sindresorhus/env-paths#pathsconfig

sindresorhus (changes requested):

sindresorhus:
You need to document it in the readme too.

ncallaway:
I'll make the changes requested, and the CI formatting fixes shortly. The only question I have is the proper destination for the `mode` link.

The node documentation for `fs.open()` (where they describe `mode` in the most detail) links to https://man7.org/linux/man-pages/man2/open.2.html . Another option would be the wikipedia page for File system permissions numeric notation section (https://en.wikipedia.org/wiki/File-system_permissions#Numeric_notation), or other more casual explainer: https://docs.nersc.gov/filesystems/unix-file-permissions/.

Do you have any preferences for the level of detail in the explanation link for the `mode` options?

ncallaway:
I've pushed a change as a separate commit that does the following:

- Renames the parameter from `mode` to `configFileMode`
- updates the type definition to only allow `number` for `configFileMode`
- updates the documentation of `mode` with more details (including a link to the wikipedia link above https://en.wikipedia.org/wiki/File-system_permissions#Numeric_notation)
- adds the mode documentation to `readme.md`
- updates the formatting to pass `npm run test`

This is a separate commit to make it easy to see the changes from before, but I'm also happy to squash into a single commit before this is merged.

PR #162, opened 2022-03-20
Feature: Allow schema to be free-form #80

dawidgarus:
Fixes https://github.com/sindresorhus/conf/issues/80

### DIFF
--- a/test/index.ts
+++ b/test/index.ts
@@ -702,6 +702,30 @@ test('schema - valid set', t => {
 	});
 });
 
+test('schema - valid set with full schema', t => {
+	const schema: Schema<{foo: {bar: number; foobar: number}}> = {
+		type: 'object',
+		properties: {
+			foo: {
+				type: 'object',
+				properties: {
+					bar: {
+						type: 'number'
+					},
+					foobar: {
+						type: 'number',
+						maximum: 100
+					}
+				}
+			}
+		}
+	};
+	const config = new Conf({cwd: tempy.directory(), schema});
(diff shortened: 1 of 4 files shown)


sindresorhus:
Can you update the readme too?

test/index.ts
```diff
@@ -702,6 +702,30 @@ test('schema - valid set', t => {
...
+					bar: {
+						type: 'number'
+					},
+					foobar: {
+						type: 'number',
+						maximum: 100
+					}
+				}
+			}
+		}
```
sindresorhus, line 721: Can you include a top-level only property? Just to make sure it works.

dawidgarus: Added additional tests

test/index.ts
```diff
@@ -702,6 +702,43 @@ test('schema - valid set', t => {
 	});
 });
 
+test('schema - valid set with full schema', t => {
+	const schema: Schema<{foo: {bar: number; foobar: number}}> = {
+		type: 'object',
```
sindresorhus, line 707: Can you throw a human-friendly error if the user tries to use any other value than `object` for the `type` property? It would help non-TS users to not incorrectly use it.

dawidgarus: Added check and additional tests for the check

sindresorhus:
https://github.com/sindresorhus/conf/pull/162#issuecomment-1074300768

^

CalebUsadi:
I would like to be assigned to the completion of this issue. Thanks!

PR #166, opened 2022-06-16
Add overloads to onDidChange method to support dot-notation access

### LINKED ISSUE #165: onDidChange doesn't support dot-notation for nested properties
I get a typescript compilation error when I try to pass in a key with dot-notation to the `onDidChange` method. This seems to happen because the type of key the onDidChange method is expecting is `keyof <StoreType>`. This doesn't include the nested object properties.

Requesting to add an overload that takes a key of type `string`. Just like the `get` or `set`.

Thanks!

```ts
interface Config {
  foo: Foo;
};

interface Foo {
  bar: string;
};

export const defaults: Config = {
  foo: {
    bar: "foobar"
  }
};

const store = new Store<Config>({
  defaults,
});

// ERROR - Argument of type '"foo.bar"' is not assignable to parameter of type 'keyof Config'.
store.onDidChange('foo.bar', (nVal, oVal) => {
  console.log(nVal, oVal);
});
```

hdhilip98:
Fixes #165

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -293,14 +293,14 @@ class Conf<T extends Record<string, any> = Record<string, unknown>> implements I
 	/**
 	Watches the given `key`, calling `callback` on any changes.
 
-	@param key - The key wo watch.
+	@param key - The key to watch.
 	@param callback - A callback function that is called on any changes. When a `key` is first set `oldValue` will be `undefined`, and when a key is deleted `newValue` will be `undefined`.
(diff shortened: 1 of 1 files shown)


sindresorhus:
This needs a type test: https://github.com/sindresorhus/conf/blob/main/test/index.test-d.ts

sindresorhus:
@hdhilip98 Bump

PR #167, opened 2022-07-14
beforeMigration option

Ciberusps:
You can use `beforeMigration` callback for logging purposes, preparing migration data, etc.
`beforeMigration` callback will be called before the migration is executed.
It could be useful when u have multiple configs and you want to log the migration data for each config.

My example - I use multiple `electron-store`'s and its usefull to log all migrations via `electron-log` in single file for debugging purposes, its easier to understand what can cause issue. Probably its can be usefull for others

```ts
import Conf from 'conf';

console.log = someLogger.log;

const mainConfig = new Conf({
	beforeMigration: (store, currentVersion, nextVersion) => {
		console.log(`[main-config] migrate from ${currentVersion} -> ${nextVersion}`);
	},
	migrations: {
		'0.4.0': store => {
		   // somethign that might broke config silently
        },
	}
});

const secondConfig = new Conf({
	beforeMigration: (store, currentVersion, nextVersion) => {
		console.log(`[second-config] migrate from ${currentVersion} -> ${nextVersion}`);
	},
	migrations: {
		'1.0.1': store => {
          // somethign that might broke config silently
        },
	}
});
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -136,11 +136,9 @@ const store = new Conf({
 Type: `Function`\
 Default: `undefined`
 
-You can use `beforeEachMigration` callback for logging purposes, preparing migration data, etc.
+The given callback function will be called before each migration step.
 
-`beforeEachMigration` callback will be called before each migrations executed.
-
-It could be useful when you have multiple configs and you want to log the migration data for each config.
+This can be useful for logging purposes, preparing migration data, etc.
 
 Example:
 
--- a/readme.md
+++ b/readme.md
@@ -159,6 +159,7 @@ const mainConfig = new Conf({
 		},
 	}
 });
+
 const secondConfig = new Conf({
 	beforeEachMigration: (store, context) => {
 		console.log(`[second-config] migrate from ${context.fromVersion} → ${context.toVersion}`);
--- a/readme.md
+++ b/readme.md
@@ -131,16 +131,16 @@ const store = new Conf({
 
 > Note: The version the migrations use refers to the **project version** by default. If you want to change this behavior, specify the [`projectVersion`](#projectVersion) option.
 
-### beforeMigration
+### beforeEachMigration
 
 Type: `Function`\
 Default: `undefined`
 
-You can use `beforeMigration` callback for logging purposes, preparing migration data, etc.
+You can use `beforeEachMigration` callback for logging purposes, preparing migration data, etc.
 
-`beforeMigration` callback will be called before the migration is executed.
+`beforeEachMigration` callback will be called before each migrations executed.
 
-It could be useful when u have multiple configs and you want to log the migration data for each config.
+It could be useful when you have multiple configs and you want to log the migration data for each config.
 
 Example:
 
@@ -150,8 +150,8 @@ const Conf = require('conf');
 console.log = someLogger.log;
 
 const mainConfig = new Conf({
-	beforeMigration: (store, currentVersion, nextVersion) => {
-		console.log(`[main-config] migrate from ${currentVersion} -> ${nextVersion}`);
+	beforeEachMigration: (store, context) => {
+		console.log(`[main-config] migrate from ${context.fromVersion} → ${context.toVersion}`);
 	},
 	migrations: {
 		'0.4.0': store => {
@@ -160,8 +160,8 @@ const mainConfig = new Conf({
 	}
 });
 const secondConfig = new Conf({
-	beforeMigration: (store, currentVersion, nextVersion) => {
-		console.log(`[second-config] migrate from ${currentVersion} -> ${nextVersion}`);
+	beforeEachMigration: (store, context) => {
+		console.log(`[second-config] migrate from ${context.fromVersion} → ${context.toVersion}`);
 	},
 	migrations: {
 		'1.0.1': store => {
--- a/readme.md
+++ b/readme.md
@@ -131,6 +131,46 @@ const store = new Conf({
 
 > Note: The version the migrations use refers to the **project version** by default. If you want to change this behavior, specify the [`projectVersion`](#projectVersion) option.
 
+### beforeMigration
+
+Type: `Function`\
+Default: `undefined`
+
+You can use `beforeMigration` callback for logging purposes, preparing migration data, etc.
+
+`beforeMigration` callback will be called before the migration is executed.
+
+It could be useful when u have multiple configs and you want to log the migration data for each config.
+
+Example:
+
+```js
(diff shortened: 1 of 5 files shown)


source/index.ts
```diff
@@ -14,11 +14,11 @@ import debounceFn = require('debounce-fn');
 import semver = require('semver');
 import onetime = require('onetime');
 import {JSONSchema} from 'json-schema-typed';
-import {Deserialize, Migrations, OnDidChangeCallback, Options, Serialize, Unsubscribe, Schema, OnDidAnyChangeCallback} from './types';
+import {Deserialize, Migrations, OnDidChangeCallback, Options, Serialize, Unsubscribe, Schema, OnDidAnyChangeCallback, BeforeMigrationCallback} from './types';
 
 const encryptionAlgorithm = 'aes-256-cbc';
 
-const createPlainObject = <T = unknown>(): T => {
+const createPlainObject = <T = Record<string, unknown>>(): T => {
```
Ciberusps, line 21: type parameter constraint for TS 4.7

readme.md
```diff
@@ -131,6 +131,46 @@ const store = new Conf({
...
+
+Example:
+
+```js
+const Conf = require('conf');
+
+console.log = someLogger.log;
+
+const mainConfig = new Conf({
+	beforeMigration: (store, currentVersion, nextVersion) => {
```
sindresorhus, line 153: I would prefer if it were like this.
```suggestion
	beforeMigration: (store, context) => {
```

That way we could easily add more metadata properties in the future.

sindresorhus: Should the properties maybe be named `fromVersion` and `toVersion`? `currentVersion` makes it sound like it's the current version of the package, not the current version in the migration process.

readme.md
```diff
@@ -131,6 +131,46 @@ const store = new Conf({
...
+### beforeMigration
+
+Type: `Function`\
+Default: `undefined`
+
+You can use `beforeMigration` callback for logging purposes, preparing migration data, etc.
+
+`beforeMigration` callback will be called before the migration is executed.
+
+It could be useful when u have multiple configs and you want to log the migration data for each config.
```
sindresorhus, line 143: Write proper English. No "u".

readme.md
```diff
@@ -131,6 +131,46 @@ const store = new Conf({
...
 > Note: The version the migrations use refers to the **project version** by default. If you want to change this behavior, specify the [`projectVersion`](#projectVersion) option.
 
+### beforeMigration
+
+Type: `Function`\
+Default: `undefined`
+
+You can use `beforeMigration` callback for logging purposes, preparing migration data, etc.
+
+`beforeMigration` callback will be called before the migration is executed.
```
sindresorhus, line 141: This is ambiguous. Before the whole process or before each step.

readme.md
```diff
@@ -131,6 +131,46 @@ const store = new Conf({
...
+const mainConfig = new Conf({
+	beforeMigration: (store, currentVersion, nextVersion) => {
+		console.log(`[main-config] migrate from ${currentVersion} -> ${nextVersion}`);
+	},
+	migrations: {
+		'0.4.0': store => {
+			store.set('debugPhase', true);
+		},
+	}
+});
```
sindresorhus, line 161:
```suggestion
});

```

test/index.ts
```diff
@@ -1203,3 +1203,22 @@ test('__internal__ keys - should only match specific "__internal__" entry', t =>
 		conf.set('__internal__foo.you-shall', 'not-pass');
 	});
 });
+
+test('beforeMigration - should be called before every migration', t => {
+	const conf = new Conf({
+		cwd: tempy.directory(),
+		projectVersion: '2.0.0',
+		beforeMigration: (store, currentVersion, nextVersion) => {
+			store.set(`beforeMigration ${currentVersion} -> ${nextVersion}`, true);
```
sindresorhus, line 1212:
```suggestion
			store.set(`beforeMigration ${currentVersion} → ${nextVersion}`, true);
```

sindresorhus (changes requested):

sindresorhus:
I think it should be called `beforeEachMigration` to make it clear it's called for each step and that would also allow adding a `beforeMigration` hook in the future which would be called only once before all migrations are done.

readme.md
Ciberusps: > We could maybe also add the package version and all the migration versions. In case the user wants to know where in the migration process they are.


Added `context.migrationProcess` with `finalVersion` and `migrations` not sure how to name them properly probably they can be flattened and lay in context but proper naming required

sindresorhus: I would prefer to keep them top-level in `context` (no nesting).

source/types.ts
```diff
@@ -103,6 +103,16 @@ export interface Options<T extends Record<string, any>> {
...
 	migrations?: Migrations<T>;
 
+	/**
+	You can use `beforeEachMigration` callback for logging purposes, preparing migration data, etc.
+
+	`beforeEachMigration` callback will be called before the migration is executed.
+
+	It could be useful when you have multiple configs and you want to log the migration data for each config.
+
+	*/
```
sindresorhus, line 113:
```suggestion
	*/
```

test/index.test-d.ts
```diff
@@ -135,6 +135,11 @@ expectError<number>(config.get('unicorn', 1));
 
 // -- Migrations --
 new Conf({
+	beforeEachMigration: (store, context) => {
+		console.log(`[main-config] migrate from ${context.fromVersion} → ${context.toVersion}`);
+		console.log(`[main-config] final migration verison ${context.migrationProcess.finalVersion}, all migrations that were run or will be runned: ${context.migrationProcess.versions.toString()}`);
```
sindresorhus, line 140:
```suggestion
		console.log(`[main-config] final migration version ${context.migrationProcess.finalVersion}, all migrations that were run or will be ran: ${context.migrationProcess.versions.toString()}`);
```

test/index.ts
```diff
@@ -1203,3 +1203,22 @@ test('__internal__ keys - should only match specific "__internal__" entry', t =>
...
+			store.set(`beforeEachMigration ${context.fromVersion} → ${context.toVersion}`, true);
+		},
+		migrations: {
+			'1.0.0': () => {},
+			'1.0.1': () => {},
+			'2.0.1': () => {}
+		}
+	});
+
+	t.is(conf.get('beforeEachMigration 0.0.0 → 1.0.0'), true);
```
sindresorhus, line 1221: Use `t.true`

PR #177, opened 2023-03-14
Fixed TypeScript TS2307 error

jculloty:
Typescript reports the following error 

TS2307: Cannot find module 'conf' or its corresponding type declarations

This is because the package.json is missing a types or main filed.

`
  "types": "./dist/source/index.d.ts",
  "main": "./dist/source/index.js",
`
Without these fields **tsc --build  --traceResolution** results in the below

> Found 'package.json' at 'C:/Users/******/node_modules/conf/package.json'.
> 'package.json' does not have a 'typesVersions' field.
> File 'C:/Users/******/node_modules/conf.ts' does not exist.
> File 'C:/Users/******/node_modules/conf.tsx' does not exist.
> File 'C:/Users/******/node_modules/conf.d.ts' does not exist.
> 'package.json' does not have a 'typings' field.
> 'package.json' does not have a 'types' field.
> 'package.json' does not have a 'main' field.

While there is an exports filed in the current package.json

`
  "exports": {
    "types": "./dist/source/index.d.ts",
    "default": "./dist/source/index.js"
  }
`

It appears this is ignored unless Typescripts  moduleResolution is updated "node16" or "nodenext".

Unfortunately this can cause issues with other imported packages. Adding these two fields will allow the package to be used with moduleResolution "node"

[For any environments supporting exports main will be ignored.](https://nodejs.org/api/packages.html#package-entry-points)

> For new packages targeting the currently supported versions of Node.js, the ["exports"](https://nodejs.org/api/packages.html#exports) field is recommended. For packages supporting Node.js 10 and below, the ["main"](https://nodejs.org/api/packages.html#main) field is required. If both ["exports"](https://nodejs.org/api/packages.html#exports) and ["main"](https://nodejs.org/api/packages.html#main) are defined, the ["exports"](https://nodejs.org/api/packages.html#exports) field takes precedence over ["main"](https://nodejs.org/api/packages.html#main) in supported versions of Node.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,8 @@
 		"types": "./dist/source/index.d.ts",
 		"default": "./dist/source/index.js"
 	},
+	"types": "./dist/source/index.d.ts",
+	"main": "./dist/source/index.js",
 	"engines": {
 		"node": ">=14.16"
 	},


sindresorhus:
This project requires `node16` and I don't intend to change that.

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

ryangoree:
> This project requires `node16` and I don't intend to change that.
> 
> https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

I don't know if you're on the same page as the first reply to that gist, but the author links to 2 package.json examples that both include a `main` file as examples of the "right" way to do it.

I'm confused about the aversion to adding adding even a `main` field?

vaaski:
> This project requires `node16` and I don't intend to change that.

Why does it say `>=14.16` then
https://github.com/sindresorhus/conf/blob/184fc278736dee34c44d4e7fa7e1b2a16ffdd5be/package.json#L19

Also I am on node 18 with `"type": "module"` and it still shows the same error. 
https://github.com/vaaski/howtf/blob/1195a310b71cfd4e3988d2c958d43d60b54649c4/src/config.ts

bokub:
> This project requires node16 and I don't intend to change that.

I'm using node 18 and I have the same problem. I had to rollback to version 10.

PR #182, opened 2023-07-17
Fix types on Conf::delete to allow access by dot notation

t3chguy:


source/index.ts
```diff
@@ -249,7 +249,10 @@ export default class Conf<T extends Record<string, any> = Record<string, unknown
 
 	@param key - The key of the item to delete.
 	*/
-	delete<Key extends keyof T>(key: Key): void {
+	delete<Key extends keyof T>(key: Key): void;
+	// This overload is used for dot-notation access.
+	delete<Key extends string>(key: Exclude<Key, keyof T>): void;
```
sindresorhus, line 254: I don't think this overload is needed.

t3chguy: Indeed

PR #184, opened 2023-09-07
Read data from disk once

fablee1:
Each read right now reads the data from the disk, which affects performance if conf is used in a high load application as the main storage.
This change stores each write additionally in the state and on each use of the store it deep clones the state and returns it instead of reading from disk. Deep cloning is needed to prevent the mutation of the state variable and ensure conf fully behaves as previously.
Speed improvements for reads is 100-500x, especially if encryption is used.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -50,6 +50,8 @@ const checkValueType = (key: string, value: unknown): void => {
 	}
 };
 
+const deepClone = <T>(obj: T): T => JSON.parse(JSON.stringify(obj));
+
 const INTERNAL_KEY = '__internal__';
 const MIGRATION_KEY = `${INTERNAL_KEY}.migrations.version`;
 
@@ -60,6 +62,7 @@ export default class Conf<T extends Record<string, any> = Record<string, unknown
 	readonly #encryptionKey?: string | Buffer | NodeJS.TypedArray | DataView;
 	readonly #options: Readonly<Partial<Options<T>>>;
 	readonly #defaultValues: Partial<T> = {};
+	#state?: T
 
 	constructor(partialOptions: Readonly<Partial<Options<T>>> = {}) {
 		const options: Partial<Options<T>> = {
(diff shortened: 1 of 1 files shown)


source/index.ts
```diff
@@ -60,6 +60,8 @@ export default class Conf<T extends Record<string, any> = Record<string, unknown
 	readonly #encryptionKey?: string | Buffer | NodeJS.TypedArray | DataView;
 	readonly #options: Readonly<Partial<Options<T>>>;
 	readonly #defaultValues: Partial<T> = {};
+	#readFromDisk = true;
+	#state?: string;
```
antongolub, line 64: I support the PR, but I'd like to suggest a few improvements, if you don't mind.

1. To avoid breaking change introduce new option like `memoize/useMemoryState`
2. 
```ts
get store(): T {
		const state = createPlainObject();

		if (this.options.memoize) {
		   if (this.#state) {
		      return this.#state;
		   }

		   this.#state = state;
		}
		
		try {
			const data = fs.readFileSync(this.path, this.#encryptionKey ? null : 'utf8');
			const dataString = this._encryptData(data);
			const deserializedData = this._deserialize(dataString);
			this._validate(deserializedData);
			
			return Object.assign(state, deserializedData);
		} catch (error: unknown) {
			if ((error as any)?.code === 'ENOENT') {
				this._ensureDirectory();
				return state;
			}
			if (this.#options.clearInvalidConfig && (error as Error).name === 'SyntaxError') {
				return state;
			}
			throw error;
		}
	}
```
3. Use `change` event to trigger `state` reset instead of `_watch`
```ts
if (options.memoize) {
  this.events.on('change', this._resetState)
}
// ...
_resetState() {
   this.#state = undefined;
}
```

antongolub: @sindresorhus,

Could you plz take a look?

sindresorhus: I agree with the suggestions here. It should be behind an option. We can make the option true by default at some point when it's more battle-tested.

PR #185, opened 2023-10-17
Add types field to package.json to fix type declarations resolution

alexanderkasten:
After updating from v10.2.0 to v11.0.2  I get the error:

Bildschirmfoto 2023-10-17 um 14 33 21

This PR fixes this bug by adding the `types` field in the package.json. This allows VSCode/TS to find the declaration files.

https://github.com/sindresorhus/conf/assets/17065920/5557d895-8e80-48cc-aa8e-b190699c834e

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
+	"types": "./dist/source/index.d.ts",
 	"exports": {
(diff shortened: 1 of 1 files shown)


sindresorhus:
The problem is the tool you are using, not `conf`. `conf` defines the type path in the `exports` property, which is a standard way to do it.

PR #189, opened 2023-10-24
feat: replace buffer with uint8array

### LINKED ISSUE #188: Switch from Buffer to Uint8Array
Hi!

I have read your [Goodbye, Node.js Buffer](https://sindresorhus.com/blog/goodbye-nodejs-buffer) post where you mentioned that you'd be open to contributions to move your packages to use `Uint8Array`.

I'd like to volunteer to do so for this package.

I have already [a fork with the changes & tests passing](https://github.com/sindresorhus/conf/compare/main...dreamorosi:conf:feat/swap_buffer_with_uint8array), I have used your [uint8array-extras](https://github.com/sindresorhus/uint8array-extras) package. 

Let me know if a PR would be welcome, if not or have different plans no worries!

dreamorosi:
Fixes #188

This PR replaces usage of `Buffer` in favor of `Uint8Array`. This is done by taking a dependency on the `uint8array-extras` package.

I verified that the tests are passing locally and I have update the documentation in the docstrings & README file.

### DIFF
--- a/source/index.ts
+++ b/source/index.ts
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/naming-convention, @typescript-eslint/no-unsafe-return, @typescript-eslint/no-redundant-type-constituents */
 import {isDeepStrictEqual} from 'node:util';
 import process from 'node:process';
-import {Buffer} from 'node:buffer';
 import fs from 'node:fs';
 import path from 'node:path';
 import crypto from 'node:crypto';
@@ -15,6 +14,12 @@ import ajvFormatsModule from 'ajv-formats';
 import debounceFn from 'debounce-fn';
 import semver from 'semver';
 import {type JSONSchema} from 'json-schema-typed';
+import {
+	concatUint8Arrays,
+	isUint8Array,
+	stringToUint8Array,
+	uint8ArrayToString,
+} from 'uint8array-extras';
 import {
 	type Deserialize,
 	type Migrations,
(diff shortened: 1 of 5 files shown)


source/index.ts
```diff
@@ -353,17 +358,19 @@ export default class Conf<T extends Record<string, any> = Record<string, unknown
 		}
 	}
 
-	private _encryptData(data: string | Buffer): string {
+	private _encryptData(data: string | Uint8Array): string {
 		if (!this.#encryptionKey) {
-			return data.toString();
+			return isUint8Array(data) ? uint8ArrayToString(data) : data.toString();
```
sindresorhus, line 363: This should be simpler and faster:

```suggestion
			return typeof data === 'string' ? string : uint8ArrayToString(data);
```

sindresorhus (changes requested):

source/index.ts
dreamorosi: Good point!

dreamorosi:
I see that the checks for Node.js 14 are failing. Looking at the error it seems that the non compatible code comes from the `uint8array-extras` package at [this line](https://github.com/sindresorhus/uint8array-extras/blob/a9095b09d9250a51658d11e50053318c209eac35/index.js#L18).

sindresorhus:
Thanks :)

sindresorhus:
> I see that the checks for Node.js 14 are failing. Looking at the error it seems that the non compatible code comes from the uint8array-extras package at [this line](https://github.com/sindresorhus/uint8array-extras/blob/a9095b09d9250a51658d11e50053318c209eac35/index.js#L18).

I'm going to require Node.js 18.

PR #190, opened 2023-11-19
Fix: cannot find module 'conf' in Typescript projects

nosajio:
I discovered that Typescript doesn't know where to look for the `conf` type declaration. The issue can be reproduced by creating a new typescript / node project, running `npm install conf`, and importing conf into a ts file without installing `@types/conf`.

Since the `@types/conf` package is now deprecated, I've added a line to the package.json to ensure the declaration file is always found in typescript projects.

image

The fix is confirmed to solve the issue by running `tsc` to see the import error, adding the types property the package file directly in `node_modules/conf`, and running tsc again to see a successful build.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
+	"types": "./dist/source/index.d.ts",
 	"exports": {
 		"types": "./dist/source/index.d.ts",
 		"default": "./dist/source/index.js"


sindresorhus:
The problem is your tsconfig, not this package. The types are defined right here: https://github.com/sindresorhus/conf/blob/ff1cc63a372823e7d68afa8a40a1e600fd36e82b/package.json#L15

https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm

PR #193, opened 2024-05-13
Fix corresponding type declarations

hywax:
Need to add a "types" field to solve the TS problem
```
TS2307: Cannot find module conf or its corresponding type declarations.
```

It solves issue https://github.com/sindresorhus/electron-store/issues/276

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
 		"url": "https://sindresorhus.com"
 	},
 	"type": "module",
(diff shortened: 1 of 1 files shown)


sindresorhus:
This is not correct. If you have configured your project correctly, TypeScript will read the types location from the `exports` field. Adding `"types"` may seem like a workaround, but it could have unintended consequences. The current package.json is correct.

PR #196, opened 2024-12-02
Add `rootSchema` and `ajvOptions` options

### LINKED ISSUE #80: Feature Request: Allow schema to be free-form

### LINKED ISSUE #99: Feacture Request: ajv should have "removeAdditional: 'all' "  option

### LINKED ISSUE #169: Ajv is in strictmode
Ajv is in strictmode which makes the usage of validation keywords impossible.

Solution: 
Disable strictmode for ajv.

CalebUsadi:
Fixes #80
Fixes #99 
Fixes #169

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -101,9 +101,10 @@ config.set('foo', '1');
 **Note:** The `default` value will be overwritten by the `defaults` option if set.
 
 #### rootSchema
+
 Type: `object`
 
-Top-level properties for the schema. Requires a `schema` option to be provided, and cannot contain a `properties` field.
+Top-level properties for the schema. Requires the `schema` option to be specified, and cannot contain a `properties` field.
 
 Example:
 
@@ -120,9 +121,10 @@ const store = new Conf({
 ```
 
 #### ajvOptions
+
 Type: `object`
 
-Options to pass to AJV. Requires a `schema` option to be provided.
+Options passed to AJV. Requires the `schema` option to be specified.
 
 Example:
 
@@ -136,8 +138,8 @@ const store = new Conf({
 		additionalProperties: false
 	},
 	ajvOptions: {
-		removeAdditional: true,
-	},
+		removeAdditional: true
+	}
 });
 ```
 
--- a/readme.md
+++ b/readme.md
@@ -100,6 +100,47 @@ config.set('foo', '1');
 
 **Note:** The `default` value will be overwritten by the `defaults` option if set.
 
+#### rootSchema
+Type: `object`
+
+Top-level properties for the schema. Requires a `schema` option to be provided, and cannot contain a `properties` field.
+
+Example:
+
+```js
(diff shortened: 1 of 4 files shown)


source/types.ts
```diff
@@ -50,6 +51,20 @@ export type Options<T extends Record<string, any>> = {
...
+
+	Requires a `schema` option to be provided, and cannot contain a `properties` field.
+	*/
+	rootSchema?: Omit<TypedJSONSchema, 'properties'>;
+
+	/**
+	Options to pass to AJV.
+
+	Requires a `schema` option to be provided.
+	*/
```
sindresorhus, line 65: This should be in sync with the readme. (missing examples.

source/types.ts
```diff
@@ -50,6 +51,20 @@ export type Options<T extends Record<string, any>> = {
...
+	Top-level properties for the schema.
+
+	Requires a `schema` option to be provided, and cannot contain a `properties` field.
+	*/
+	rootSchema?: Omit<TypedJSONSchema, 'properties'>;
+
+	/**
+	Options to pass to AJV.
+
+	Requires a `schema` option to be provided.
```
sindresorhus, line 64:
```suggestion
	Requires the `schema` option to be specified.
```

readme.md
```diff
@@ -100,6 +100,49 @@ config.set('foo', '1');
...
+		additionalProperties: false
+	}
+});
+```
+
+#### ajvOptions
+
+Type: `object`
+
+Options passed to AJV. Requires the `schema` option to be specified.
```
sindresorhus, line 127: Linkify `Options passed to AJV.` to the AJV options docs.

sindresorhus (changes requested):

sindresorhus:
The pull request needs a proper title.

PR #198, opened 2025-05-18
feat: add `.getAll()` to get all config items

wy-luke:
Assuming that I have a config like below:

```json
{
  "a": "a",
  "b": "b",
  "c": "c",
  "d": "d"
}
```

To get all configs, I must use `get(xx)` four times. I think maybe it's a good idea to use `get()` to retrieve all configs at one time.
When using `electron`, it can also get all configs with only one disk IO.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -25,8 +25,8 @@ const config = new Conf({projectName: 'foo'});
 config.set('unicorn', '🦄');
 console.log(config.get('unicorn'));
 //=> '🦄'
-console.log(config.getAll({unicorn: '🦄'}));
-//=> {unicorn: '🦄', foo: {bar: true}}
+console.log(config.getAll());
+//=> {projectName: 'foo', unicorn: '🦄'}
 
 // Use dot-notation to access nested properties
 config.set('foo.bar', true);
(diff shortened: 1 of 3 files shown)


sindresorhus:
This should be a separate method.

wy-luke:
@sindresorhus  How about this: 7019c7c8efbb380d1d7704eaea6272690e59649f

readme.md
```diff
@@ -405,6 +407,10 @@ Set multiple items at once.
 
 Get an item or `defaultValue` if the item does not exist.
 
+#### .getAll(defaultValue?)
+
+Get all config items or `defaultValue` if no items exist.
```
sindresorhus, line 412: We don't need default value here. Can just return an empty object if no items exist.

sindresorhus:
Needs a test.

readme.md
wy-luke: @sindresorhus Done.

wy-luke:
@sindresorhus All done. Take a look plz.

sindresorhus:
I just realized, this already exists as `config.store`.

wy-luke:
> I just realized, this already exists as `config.store`.
I think getAll would be more clear and easy to know. Besides, getAll can return {} when config.store is null or undefined, right?

But still, it's up to you. ❤️

sindresorhus:
> Besides, getAll can return {} when config.store is null or undefined, right?

No, it always returns an object.

wy-luke:
@sindresorhus At least, adding this usage to the Readme is necessary. It's easy to use but too hard to know it, even you realize its existence later. 😂

By the way, I wonder why you don't accept passing `undefined` to `.get()`. Is there any consideration in terms of safety? Thanks!

PR #208, opened 2026-03-27
feat: add deserializeComplexTypes option for Date round-tripping

### LINKED ISSUE #18: Support dates
Would be nice if you could `store.set('foo', new Date())` and get back a `Date` object when you do `store.get('foo')`.

phusi319:
Closes https://github.com/sindresorhus/electron-store/issues/18

## Summary

Adds a new `deserializeComplexTypes` option that preserves `Date` objects through JSON serialization/deserialization. This enables:

`js
const config = new Conf({
  projectName: 'foo',
  deserializeComplexTypes: true
});

config.set('timestamp', new Date());
config.get('timestamp') instanceof Date;
//=> true
`

## Approach

Uses JSON's built-in replacer/reviver mechanism (as suggested in the original issue) with a tagged wrapper format:

`json
{
  "timestamp": {
    "`___BEGIN___COMMAND_DONE_MARKER___$LASTEXITCODE`type": "Date",
    "`___BEGIN___COMMAND_DONE_MARKER___$LASTEXITCODE`value": "2024-06-15T12:00:00.000Z"
  }
}
`

### Why this approach?

- **No separate metadata tracking**  type info is stored inline, so nested objects, arrays, and dot-prop access all work automatically without fragile path-tracking
- **Collision-safe**  the reviver only matches objects with *exactly* two keys (`___BEGIN___COMMAND_DONE_MARKER___$LASTEXITCODEtype` and `___BEGIN___COMMAND_DONE_MARKER___$LASTEXITCODEvalue`) and a known type string. Objects with extra keys or unknown types pass through unchanged
- **Backward compatible**  opt-in via `deserializeComplexTypes: true`. Existing configs are unaffected
- **Extensible**  new types (Map, Set, RegExp, etc.) can be added by extending the replacer/reviver without changing the format

### Design decisions

- **Opt-in**  avoids breaking existing configs. Users explicitly enable the feature
- **Ignored with custom serialize/deserialize**  if users provide their own serialization, they control the format
- **Uses `this[key]` in replacer**  accesses the raw Date before `.toJSON()` converts it to a string

## What's included

- `deserializeComplexTypes` option in types and implementation
- JSON replacer (`complexTypeReplacer`) and reviver (`complexTypeReviver`)
- Enhanced `#configureSerialization` with clean fallback logic
- Comprehensive test suite (`test/complex-types.ts`):
  - Basic Date round-tripping
  - Dot-notation keys
  - Nested objects with multiple Date fields
  - Arrays (pure Date and mixed)
  - Deeply nested structures (3+ levels)
  - Invalid Date handling
  - False positive prevention (extra keys, missing keys, unknown types)
  - Raw JSON file format verification
  - `.store` getter/setter
  - Encryption compatibility
  - Custom serialize/deserialize override behavior
  - `accessPropertiesByDotNotation: false` compatibility
- README documentation with example
- Full JSDoc with example in types.ts

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -420,6 +420,34 @@ You would usually not need this, but it could be useful if you use a custom `cwd
 > [!NOTE]
 > Setting restrictive permissions can cause problems if different users need to read the file. A common problem is a user running your tool with and without `sudo` and then not being able to access the config the second time.
 
+#### deserializeComplexTypes
+
+Type: `boolean`\
+Default: `false`
+
+Preserve non-JSON types like `Date` through serialization and deserialization.
+
+When enabled, `Date` objects are tagged during serialization so they can be restored as proper `Date` instances when read back. This allows round-tripping of `Date` values.
+
+```js
+import Conf from 'conf';
+
+const config = new Conf({
+	projectName: 'foo',
+	deserializeComplexTypes: true
+});
+
+config.set('timestamp', new Date());
+console.log(config.get('timestamp') instanceof Date);
+//=> true
+
+config.set('nested', {createdAt: new Date()});
+console.log(config.get('nested').createdAt instanceof Date);
+//=> true
+```
+
+This option is ignored when custom `serialize` or `deserialize` functions are provided.
+
 ### Instance
 
 You can use [dot-notation](https://github.com/sindresorhus/dot-prop) in a `key` to access nested properties.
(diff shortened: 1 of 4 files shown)


sindresorhus:
Thanks, but I already declined this in https://github.com/sindresorhus/conf/pull/154#issuecomment-3288627561

